PR #18, opened 2025-05-11
Feat: Remove test directory restriction from WriteTool

endoze:
In order to enable more use cases for the WriteTool, this commit removes the `/tests` restriction on the WriteTool. It adds a configurable parameter that allows restricting where the WriteTool can actually write files to so we can preserve the existing behavior through configuration.


parruda (approved):
Hey @endoze this is great.

PR #27, opened 2025-05-14
[dont merge] Add rubocop violation

jonallured:
Sibling PR to https://github.com/Shopify/roast/pull/26 where I'm experimenting with our CI setup - this should fail CI!

```
$ bundle exec rubocop; echo "last exit was" $?
Inspecting 66 files
..........C.......................................................

Offenses:

lib/roast.rb:14:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
lib/roast.rb:15:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
lib/roast.rb:16:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
lib/roast.rb:17:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
lib/roast.rb:18:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.

66 files inspected, 5 offenses detected, 5 offenses autocorrectable
last exit was 1
```

PR #28, opened 2025-05-14
Remove RuboCop autocorrect from default rake task

jonallured:
While the autocorrect flag is convienent for the default Rake task it has the unintended side effect of causing RuboCop violations in CI to be corrected instead of failing a PR! This PR just removes the flag so that we can rest assured that CI will fail as expected. 😅 

See also: https://github.com/Shopify/roast/pull/27

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,8 +12,6 @@ end
 
 task test: [:minitest]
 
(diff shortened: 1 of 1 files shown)


jonallured:
Closing in favor of #29

PR #38, opened 2025-05-15
Feat: Add support for OpenRouter

### LINKED ISSUE #21: Add suport for OpenRouter
Currently, Roast just support OpenAI client, it would be nice if it could also provide support for OpenRouter, specially given that `Raix` provides support for it.

Most probably the interesting parts of the code to dig into would be around [this section](https://github.com/Shopify/roast/blob/fd4fe2780b33bb5a1bacb67639874a7ea7147e78/lib/roast/workflow/configuration_parser.rb#L116-L140)

endoze:
In order to enable more than just OpenAI as a backend, this commit adds a new api_provider configuration option. This allows us to set openrouter or openai as the provider and correctly set up our model through Raix. It's worth mentioning that the default behavior of OpenAI is preserved and this configuration option should be only additive.

This closes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,8 +298,26 @@ Individual steps can override this setting with their own model parameter:
 
 ```yaml
 analyze_data:
-  model: anthropic:claude-3-haiku  # Takes precedence over the global model
+  model: anthropic/claude-3-haiku  # Takes precedence over the global model
 ```
+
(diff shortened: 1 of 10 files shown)


obie:
Looks great @endoze, thank you very much!

PR #43, opened 2025-05-16
Fix model access errors unhelpful

### LINKED ISSUE #13: Model access errors are not helpful
If you don't have access to a model OpenAI responds with a 404. The detailed error contains the useful parts, which are not surfaced:

```
:body => { "error" => { "message" => "The model anthropic:claude-3-7-sonnet does not exist or you do not have access to it ..."
```

This is the default state with a free OpenAI account, so a lot of people will run into this.

endoze:
In order to make errors more descriptive when things go wrong, this commit adds a few error types to handle api interactions more gracefully when things go wrong. Specifically, it currently handles when a model can't be found or is innaccesible with the api token provided. Also when an api token isn't provided, instead of an unhelpful NoMethodError, we let end users know the token wasn't provided or isn't valid.

This resolves #13

### DIFF
--- a/lib/roast/workflow/base_workflow.rb
+++ b/lib/roast/workflow/base_workflow.rb
@@ -105,17 +105,15 @@ module Roast
           })
           result
         end
+      rescue Faraday::ResourceNotFound => e
+        execution_time = Time.now - start_time
+        message = e.response.dig(:body, "error", "message") || e.message
+        error = Roast::ResourceNotFoundError.new(message)
+        e.set_backtrace([])
+        log_and_raise_error(error, message, step_model, kwargs, execution_time)
       rescue => e
         execution_time = Time.now - start_time
-
-        ActiveSupport::Notifications.instrument("roast.chat_completion.error", {
-          error: e.class.name,
-          message: e.message,
-          model: step_model,
-          parameters: kwargs.except(:openai, :model),
-          execution_time: execution_time,
-        })
-        raise
+        log_and_raise_error(e, e.message, step_model, kwargs, execution_time)
       end
 
       def workflow
@@ -123,6 +121,18 @@ module Roast
       end
 
       private
+
+      def log_and_raise_error(error, message, model, params, execution_time)
+        ActiveSupport::Notifications.instrument("roast.chat_completion.error", {
+          error: error.class.name,
+          message: message,
+          model: model,
+          parameters: params.except(:openai, :model),
+          execution_time: execution_time,
+        })
+
+        raise error
+      end
 
       # Determine the directory where the actual class is defined, not BaseWorkflow
       def determine_context_path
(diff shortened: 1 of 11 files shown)


endoze:
@dblock Sorry about that. I was unsure of how best to stub/mock in order to trigger these specific errors but I think I've got it figured out now. Have a look and let me know what you think.

lib/roast/workflow/base_workflow.rb
```diff
@@ -87,20 +87,40 @@ def chat_completion(**kwargs)
...
+      rescue Faraday::ResourceNotFound => e
+        execution_time = Time.now - start_time
+        message = e.response.dig(:body, "error", "message") || e.message
+        error = Roast::ResourceNotFoundError.new(message)
+        e.set_backtrace([])
+        log_and_raise_error(error, message, model, kwargs, execution_time)
+      rescue NoMethodError => e
+        execution_time = Time.now - start_time
+
+        if e.message.include?("undefined method 'chat' for nil")
```
dblock, line 99: Can we catch this one earlier? IMO it should never get here.

dblock (commented):
This is looking good. 

I really dislike catching `NoMethodError` though - we should prevent this from happening in code before the error happens. Why are we ever calling `.chat` on a client that could not be initialized?

lib/roast/workflow/base_workflow.rb
endoze: I've moved the logic to much earlier in the code for catching this sort of misconfiguration.

dblock:
Looks a lot better! I think some test paths are missing around raised `OpenRouter::ConfigurationError` for example? Otherwise LGTM, @obie @parruda should check this out too.

lib/roast/workflow/configuration_parser.rb
```diff
@@ -103,29 +103,37 @@ def load_roast_initializers
...
+        error = Roast::AuthenticationError.new("API authentication failed: No API token provided or token is invalid")
+        e.set_backtrace([])
+
+        ActiveSupport::Notifications.instrument("roast.workflow.start.error", {
+          error: error.class.name,
+          message: error.message,
+        })
+
+        raise error
+      rescue => e
```
dblock, line 135: In general we only rescue `StandardError => `

dblock (approved):
This LGTM, but I'd want @obie or @parruda to review/merge cause I don't know what I am doing.

obie:
Hi @endoze\! 👋 

Thank you so much for this contribution\! Your implementation for better error handling was exactly what we needed to fix issue #13.

Since this PR was based on an older version of our codebase (before some significant refactoring), I've taken your excellent work and updated it to be compatible with our current architecture in PR #82, which has now been merged.

Your original implementation has been preserved with full credit as co-author on the commit. The core improvements you made are all there:
- Custom error classes for better error handling
- Extracting meaningful error messages from API responses
- Early validation of API tokens

I'm closing this PR now that the functionality has been merged, but please know that your contribution is very much appreciated and is now part of Roast\! 🎉

Thanks again for helping make Roast better\!

PR #57, opened 2025-05-20
Targetless fixes

palkan:
There are some issues I've encountered while playing with a targetless workflow:

- `null` system prompt:

```
Chat completion failed!!!!!!!!!!!!!!!!: {"error"=>{"message"=>"Invalid value for 'content': expected a string, got null.", "type"=>"invalid_request_error", "param"=>"messages.[0].content", "code"=>nil}}
```

That happens when no workflow prompt exists and no target specified.

- Saving state for targetless executions failed because there is no `file_path`

Fixed both in this PR.


obie:
thank you @palkan

PR #59, opened 2025-05-20
Fix saving status for path/to/step.rb steps

palkan:
Steps can be stored in nested folders and named as `path/to/step.rb`, so we must create intermediate folder when saving the status.


obie:
Agree with @dblock this one needs test coverage.

PR #60, opened 2025-05-20
Fix plain text prompts configuration

palkan:
This fixes the bug with the model not being passed to the prompt step (so, the default `anthropic:claude-sonnet-3.7` is always used by text steps).

The `auto_loop: false` seems unnecessary since we set `auto_loop: false` in the PromptStep explicitly.

I decided to use the same `#step_setup` method so we can do all the configuration tweaks uniformly.


obie:
Also needs test coverage.

PR #64, opened 2025-05-20
Add workflow iteration support

obie:
## Summary
- Adds RepeatStep and EachStep to support iteration patterns in workflows
- Implements base iteration functionality that handles step execution and state management
- Includes complete tests and example workflows demonstrating the new features

## Details
This PR implements iteration support for Roast workflows through two main constructs:

1. **Repeat Step**: Executes a sequence of steps repeatedly until a condition is met or maximum iterations are reached
2. **Each Step**: Iterates through items in a collection and executes steps for each item with the current item available as a variable

These new features enable more sophisticated workflow patterns like:
- Processing collections of items (files, records, etc.)
- Implementing retry logic with conditional termination
- Building multi-step iterative processes

## Test plan
- Unit tests for RepeatStep and EachStep covering various scenarios
- Example workflows in examples/iteration/ directory
- The iteration functionality can be tested by running the simple_workflow.yml example

## Future improvements
- In a future PR, we'll standardize all Ruby expression evaluation to use the {{}} syntax consistently
- This will make it clearer when code is being evaluated vs when values are being passed literally

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,6 +30,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - Add runtime dependencies to `roast.gemspec`.
 - Add development dependencies to `Gemfile`.
 - Don't ever test private methods directly. Specs should test behavior, not implementation.
+- I do not like test-specific code embedded in production code, don't ever do that
 
 ## Git Workflow Practices
 
@@ -141,3 +142,4 @@ gh pr diff {pr_number}
     - Avoid premature optimization outside of hot paths
     - Consider the tradeoff between readability and performance
     - Suggest optimizations that improve both clarity and performance
+```
--- a/examples/grading/workflow.yml
+++ b/examples/grading/workflow.yml
@@ -1,10 +1,11 @@
-name: Grading current test changes
+name: Test Grading
 
 tools:
   - Roast::Tools::Grep
   - Roast::Tools::ReadFile
   - Roast::Tools::SearchFile
 
+# Uncomment this to run the workflow on modified tests automatically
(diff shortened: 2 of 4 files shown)


obie:
For both `each` and `until` parameters, I want to accept four types of arguments with different processing:

  1. Ruby expressions in {{}}:
    - Must be explicitly wrapped in {{}} to identify as Ruby code
    - For each: Expression should return an iterable collection
    - For until: Expression should be coerced to boolean (using !!)
  2. Bash expressions in $():
    - Command execution wrapped in $()
    - For each: The output string would be split into an iterable collection (like lines)
    - For until: Result coerced to boolean based on exit code (success=true, failure=false)
  3. Step name (string):
    - A bare string without {{}} or $() would be treated as a step name
    - For each: The step should return something iterable
    - For until: The step's result would be coerced to boolean
  4. Prompt content:
    - Similar to standard step handling
    - For each: The prompt result should be iterable
    - For until: The prompt result would be coerced to boolean

lib/roast/workflow/workflow_executor.rb
```diff
@@ -204,12 +262,88 @@ def strip_and_execute(step)
 
           # NOTE: We don't need to call interpolate here as it's already been done
           # in execute_string_step before this method is called
-          %x(#{command})
+          begin
+            output = %x(#{command})
+            raise CommandExecutionError.new("Command exited with non-zero status", step_name: command) unless $CHILD_STATUS.success?
```
dersam, line 267: This should be configurable on the step. I have some commands that I execute so I can get the failure output as context for what the workflow should do next (example: run a failing typecheck to get the output of "these are the problems to fix"). 

In those cases, I don't care if the command has a successful exit code and want to continue regardless.

PR #65, opened 2025-05-21
Allow initializer to set Raix client

jonallured:
As of https://github.com/Shopify/roast/pull/56 we had a little regression where an initializer that set the Raix api client would get clobbered by the `configure_api_client` method because the guard was no longer short circuiting. What this PR does is establish in a test that when `Raix.configuration` has either an `openrouter_client` or an `openai_client` then we should bail on the `configure_api_client` method.

I did this by checking the state of the `Raix.configuration` but I also opened https://github.com/OlympiaAI/raix/pull/25 because it would be cool for this configuration object to answer the question of whether it has a client. LMK if there's a preferences on whether to merge this as-is or to merge that and then use that solution instead!

### DIFF
--- a/lib/roast/workflow/configuration_parser.rb
+++ b/lib/roast/workflow/configuration_parser.rb
@@ -105,6 +105,9 @@ module Roast
       def configure_api_client
         return unless configuration.api_token
 
+        existing_client = !!(Raix.configuration.openrouter_client || Raix.configuration.openai_client)
+        return if existing_client
+
         begin
           case configuration.api_provider
           when :openrouter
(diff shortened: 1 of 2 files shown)


obie:
Hi @jonallured, thank you for catching this regression and taking the time to submit a fix\! 🙏

Good news - this issue has already been addressed in the current codebase. The `WorkflowInitializer` now includes a check via the `api_client_already_configured?` method that prevents overwriting existing API clients set by initializers. This achieves the same goal as your PR but with a more structured, provider-specific approach.

You can see the implementation in `lib/roast/workflow/workflow_initializer.rb` lines 39 and 73-82.

Thanks again for your contribution and for helping make Roast better\! Your vigilance in catching this regression was much appreciated.

PR #68, opened 2025-05-21
Add pause flag

### LINKED ISSUE #16: Pause at step command-line option
Add a `--pause step_name` command-line option that will stop execution with `binding.irb` after the specified step.

jonallured:
This PR adds support for a `pause` flag that takes a step name. When that step name comes up it triggers a `binding.irb` session. Fun fact prior to this PR that location had a local variable name called `step` - can you guess what happened when attempting to reference that var? That's right it.....stepped to the next line. 😝 

So in order to mitigate that I also reworked the locals in that method so that there were no collisions. 

Paired on this with @nfgrep 🍐 

Closes #16

### DIFF
--- a/lib/roast/workflow/workflow_executor.rb
+++ b/lib/roast/workflow/workflow_executor.rb
@@ -27,6 +27,7 @@ module Roast
             execute_parallel_steps(step)
           when String
             execute_string_step(step)
+            Kernel.binding.irb if workflow.pause_step_name == step # rubocop:disable Lint/Debugger
           else
             raise "Unknown step type: #{step.inspect}"
           end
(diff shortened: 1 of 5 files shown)


lib/roast/workflow/workflow_executor.rb
```diff
@@ -18,17 +18,18 @@ def initialize(workflow, config_hash, context_path)
...
           when Hash
-            execute_hash_step(step)
+            execute_hash_step(workflow_step)
           when Array
-            execute_parallel_steps(step)
+            execute_parallel_steps(workflow_step)
           when String
-            execute_string_step(step)
+            execute_string_step(workflow_step)
+            Kernel.binding.irb if workflow.pause_step_name == step
```
jonallured, line 30: Hey @nfgrep this used to be extracted to a separate method but....then the context wasn't at this code location and the `step` local wasn't in scope. Move this back and then was able to wrangle Minitest as seen in that file below.

PR #99, opened 2025-05-29
Allow configurable Cmd commands to be treated as functions (tools)

rsperko:
## Replace generic `cmd` function with command-specific functions

This PR refactors `Roast::Tools::Cmd` to register individual command functions instead of a single generic `cmd` function.

### What changed
- Commands are now individual functions: `ls()`, `pwd()`, `rake()`, etc.
- Configure allowed commands in workflows:
  ```yaml
  tools:
    - Roast::Tools::Cmd:
        allowed_commands:
          - ls
          - pwd
          - name: echo
            description: "echo command - output text to stdout"
  ```
- Default commands: `pwd`, `find`, `ls`, `rake`, `ruby`, `dev`, `mkdir`

### Why
- Better discoverability - LLMs see available commands as distinct functions
- Improved safety - only explicitly allowed commands can run
- Clearer documentation - each command has its own description

### Breaking change ⚠️
The generic `cmd` function is removed. Update workflows from:
- `cmd("ls -la")` → `ls(args: "-la")`


examples/cmd/README.md
```diff
@@ -0,0 +1,99 @@
...
+          description: "Docker CLI - container platform (e.g., docker ps, docker run)"
+```
+These descriptions help the AI understand the purpose of each command and its common subcommands or arguments.
+
+## How It Works
+
+When the `Cmd` tool is configured, each entry in `allowed_commands` is registered as a function available to the AI.
+
+- `pwd` becomes `pwd()`
+- `ls` becomes `ls(args: <string>)`
```
rsperko, line 56: @obie as this is, it IS a breaking change because `cmd(` is removed. I can think of a couple ways around that, including tying this functionality to whether `allowed_commands` is provided. I mentioned this when we talked but want to call it out here.

PR #104, opened 2025-05-29
Add support for MCP tools

macournoyer:
Usage in `workflow.yml`:

```yml
tools:
  # Load toold from an SSE MCP w/ a `url`
  - Roast Docs:
      url: https://gitmcp.io/Shopify/roast/docs
      # Can pass headers for authentication, or anything
      # env:
      #   - "Authorization: Bearer {{resource.api_token}}"
  # stdio MCPs are also supported, by providing a `command`
  - GitHub:
     command: docker
     args: ["run", "-i", "--rm", "-e", "GITHUB_PERSONAL_ACCESS_TOKEN", "ghcr.io/github/github-mcp-server"]
     env:
       GITHUB_PERSONAL_ACCESS_TOKEN: "<YOUR_TOKEN>"
     only:
       - get_issue
       - get_issue_comments
     except: # Doesn't make sense w/ only, but as an example
       - create_issue
```

I tried to keep config format as close to the MCP settings of Cursor, Claude Code and others.

Depends on: https://github.com/OlympiaAI/raix/pull/28

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,6 +749,64 @@ coding_agent(
 - Useful for tasks that require deep code understanding or multi-step changes
 - Can work across multiple files and languages
 
+### MCP (Model Context Protocol) Tools
+
+Roast supports MCP tools, allowing you to integrate external services and tools through the Model Context Protocol standard. MCP enables seamless connections to databases, APIs, and specialized tools.
+
+#### Configuring MCP Tools
+
+MCP tools are configured in the `tools` section of your workflow YAML alongside traditional Roast tools:
+
+```yaml
+tools:
+  # Traditional Roast tools
+  - Roast::Tools::ReadFile
+  
+  # MCP tools with SSE (Server-Sent Events)
+  - Documentation:
+      url: https://gitmcp.io/myorg/myrepo/docs
+      env:
+        - "Authorization: Bearer {{env.API_TOKEN}}"
+  
+  # MCP tools with stdio
+  - GitHub:
+      command: npx
+      args: ["-y", "@modelcontextprotocol/server-github"]
+      env:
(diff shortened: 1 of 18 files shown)


Gemfile
```diff
@@ -17,3 +17,5 @@ gem "rake", require: false
 gem "rubocop-shopify", require: false
 gem "vcr", require: false
 gem "webmock", require: false
+
+gem "raix", github: "macournoyer/raix", branch: "mcp-hell-yeah"
```
macournoyer, line 21: Remove once raix PR is merged

obie:
Originally in PR 104, the author separated tools into two categories:
  - local_tools - Ruby modules included directly (like Roast::Tools::Grep, Roast::Tools::ReadFile)
  - mcp_tools - MCP (Model Context Protocol) tools that connect to external services

  However, when I merged with the main branch, I found that main had already evolved the tool system to:
  - tools - An array of all tools (both strings for Ruby modules and hashes for configured tools)
  - tool_configs - A hash storing configurations for specific tools
  - Added support for tools that need post-configuration setup

  The main branch's approach is more unified - it treats all tools the same way in the tools array, whether they're simple Ruby modules or tools with configurations. This is cleaner than
  having separate local_tools and mcp_tools arrays.

  So in the merge, I kept:
  - tools from main (instead of local_tools) for regular Ruby tool modules
  - tool_configs from main for tool-specific configurations
  - mcp_tools from the PR as a separate array specifically for MCP protocol tools
  - The MCP tools are special because they need client objects and have different initialization

  This gives us the best of both worlds - the cleaner unified tool handling from main, plus the new MCP tool support from the PR.

Issue #93, opened 2025-05-29
grading example logging errors  about 'no implicit conversion of nil into String' when running Roast::Tools::SearchFile


technicalpickles:
After cloning this repo and bundling, and then running `./exe/roast execute examples/grading/workflow.yml test/roast/resources_test.rb` mentioned in the README, I get:

```
❯ DEBUG=1 ./exe/roast execute examples/grading/workflow.yml test/roast/resources_test.rb
🔥🔥🔥 Everyone loves a good roast 🔥🔥🔥

Configuring OpenAI client with token from workflow
Starting workflow...
Workflow: /Users/josh.nichols/workspace/roast/examples/grading/workflow.yml
Options: {}
Running workflow for file: test/roast/resources_test.rb
Executing: read_dependencies (Resource type: none)
[2025-05-29 11:03:32] ERROR: Error searching for 'roast/resources.rb' in '': no implicit conversion of nil into String
[2025-05-29 11:03:33] ERROR: Error searching for '**/resources.rb' in '': no implicit conversion of nil into String
[2025-05-29 11:03:34] ERROR: Error searching for 'resources.rb' in '': no implicit conversion of nil into String
```

It does eventually succeed, but the errors made me wonder what was happening.

 I tried running with `DEBUG=1` based on seeing https://github.com/Shopify/roast/blob/628d9a2f807ebeec29d7cf3581a1ac619785a9f7/lib/roast/tools/read_file.rb#L45 to log output, but it doesn't seem to give anything extra.

technicalpickles:
I switched the debug log statement to always log, and got this stack trace out of it:

```
[2025-05-29 11:33:07] ERROR: Error searching for 'roast/resources.rb' in '': no implicit conversion of nil into String
[2025-05-29 11:33:07] ERROR: /Users/josh.nichols/workspace/roast/lib/roast/tools/search_file.rb:29:in 'File.expand_path'
/Users/josh.nichols/workspace/roast/lib/roast/tools/search_file.rb:29:in 'Roast::Tools::SearchFile#call'
/Users/josh.nichols/workspace/roast/lib/roast/tools/search_file.rb:20:in 'block (2 levels) in Roast::Tools::SearchFile.included'
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:75:in 'BasicObject#instance_exec'
[...]

PR #111, opened 2025-05-30
Working grading example

technicalpickles:
Fix for https://github.com/Shopify/roast/issues/93


dblock:
Maybe this is a good opportunity to add an integration test? We could probably figure out AI access for it.

PR #115, opened 2025-05-31
Switch from arm64-darwin-23 to arm64-darwin in Gemfile.lock

technicalpickles:
I don't know at what point it was added, but bundler supports arm64-darwin without a version. This avoids needing to update explicitly or have to ignore new versions showing up after an upgrade of macOS.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,7 +164,7 @@ GEM
       hashdiff (>= 0.4.0, < 2.0.0)
 
 PLATFORMS
-  arm64-darwin-23
+  arm64-darwin
   x86_64-linux
 
 DEPENDENCIES


technicalpickles:
I'm not sure I follow. To be clear, this doesn't change what is supported in terms of platforms for gems. This doesn't affect the existing Gemfile.lock platforms for gems.  For example, this was already in the Gemfile.lock and doesn't change by applying this PR:


```
    ffi (1.17.2)
    ffi (1.17.2-arm64-darwin)
    ffi (1.17.2-x86_64-linux-gnu)
```

It just is less hassle to keep updating each revision of darwin. I'm on Sequoia 15.5 and bundling keeps updating platforms to:

```
PLATFORMS
  arm64-darwin-23
  arm64-darwin-24
  x86_64-linux
```

ericproulx:
Of course, what I met is I'm not sure the change is worth it. It would be preferable to keep just `ruby` as the platform instead of specifying specific versions. At least it would support any platforms since C extension will be compiled when `bundle install`.

technicalpickles:
This about the project's gemfile.lock, which is only used for development of the project itself.

Using prebuilt packages is generally better for developer experience because, because without compiling:
- it's faster
- it doesn't require development packages
- there's less chance for failing ti build because it's not building

I will need to confirm, but I'm pretty sure 'bundle gem' will includes specific platforms and NOT ruby, and the Darwin platforms are using the versionlees names (ie what I'm changing it to)

obie:
thanks!

PR #134, opened 2025-06-03
Refactor and use a single source .roast config dir

nfgrep:
- Added a higher level `Roast::Config` which currently just has a `root` method
- Factored initializers and cache out into their own bits of Roast::Config


nfgrep:
@nfgrep progress?


Yup! Core functionality works, need to polish and publish out of draft now.

lib/roast/dot_roast.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require "fileutils"
+require "roast/helpers/logger"
+
+module Roast
+  module DotRoast
+    class << self
+      def root(starting_path = Dir.pwd, ending_path = File.dirname(Dir.home))
+        unless starting_path.start_with?(ending_path)
```
nfgrep, line 10: The lack of this was a bug in the prev code, pwd wasn't under HOME, you'd end up hanging/looping forever.

lib/roast/helpers/function_cache/interceptor.rb
```diff
@@ -0,0 +1,90 @@
...
+                # Direct boolean value
+                cache_config
+              end
+            else
+              false
+            end
+
+            if cache_enabled
+              # Call the original function and pass in the cache
+              cache = Roast::Helpers::FunctionCache.for_workflow(workflow_configuration.name, workflow_configuration.workflow_path)
```
nfgrep, line 46: We used `configuration` in the prev code instead of `workflow_configuration` which was silently failing to do any function caching, this fixes that.

PR #138, opened 2025-06-03
Fix tool response display issue with print_response

### LINKED ISSUE #137: Tool call results displayed instead of AI response when print_response is true
## Bug Description

When using MCP tools or function calls with `print_response: true` in a workflow step, the tool call results are displayed to the user instead of the final AI response. This happens because the `PromptStep` processes and displays the intermediate array of tool results returned by `chat_completion`, rather than waiting for the final assistant response after the tool calls are processed.

## Expected Behavior

When `print_response: true` and tool calls are made:
1. The AI should make tool calls
2. The tools should execute and return results
3. The AI should process those results and generate a final response
4. **Only the final AI response should be displayed to the user**

## Actual Behavior

When `print_response: true` and tool calls are made:
1. The AI makes tool calls
2. The tools execute and return results (as an array)
3. **The tool results array is joined with newlines and displayed to the user**
4. The final AI response is generated but not displayed

## Root Cause

The issue is in `lib/roast/workflow/prompt_step.rb` lines 24-31. When `chat_completion` returns an array (which happens when tool calls are made), the code joins these tool results and passes them to `process_output`, which then displays them when `print_response` is true.

The Raix gem's `FunctionDispatch` module is supposed to handle the looping when `loop: true` is passed, but the initial call to `chat_completion` still returns the tool results array, which is what gets displayed.

## Test Case

A test has been added in `test/roast/workflow/prompt_step_tool_response_test.rb` that demonstrates this issue.

## Impact

This affects any workflow that:
- Uses MCP tools or function dispatch
- Has steps with `print_response: true`
- Makes tool calls during execution

Users see technical tool responses instead of the natural language responses from the AI.

## Potential Solutions

1. **Detect tool responses**: When `auto_loop` is true and we receive an array response, check if these are tool results and skip displaying them
2. **Modify FunctionDispatch behavior**: Change how the Raix gem handles the return value when `loop: true` to return the final response instead of intermediate results
3. **Add a flag**: Add a configuration option to control whether tool results should be displayed

Co-Authored-By: Claude <noreply@anthropic.com>

obie:
## Summary

This PR fixes an issue where `print_response: true` would display raw tool output instead of the AI's properly formatted response when tools/functions were called.

## Problem

When workflows used MCP tools or function dispatch with `print_response: true`, users would see the raw tool results (e.g., full directory listings) instead of the AI's natural language response that incorporated those results.

## Solution  

Modified `BaseStep#chat_completion` to:
- Always call with `loop: false` initially to get tool results
- Check if the response is an array (indicating tool results)
- Make a second call to get the AI's final response when needed

This ensures users see the AI's formatted response (e.g., "Here are the first 10 items...") rather than raw tool output.

## Testing

- Added test demonstrating the bug (`test/roast/workflow/prompt_step_tool_response_test.rb`)
- Updated existing tests to match new behavior
- Manually tested with MCP examples to verify correct output display

## Example

Before: Users would see raw tool output like 66 files from a directory listing
After: Users see the AI's response like "Here are the first 10 items in /Users/obie..."

Fixes #137

### DIFF
--- a/examples/grading/read_dependencies/prompt.md
+++ b/examples/grading/read_dependencies/prompt.md
@@ -5,10 +5,12 @@ The first dependency you should always look for is the source file for the prime
 If you can identify other important application-level dependencies then read them too.
 How many extra dependencies to research is left to your discretion, but ALWAYS make sure you have the subject under test (SUT) in your context before responding.
 
-Once you are finished using tool functions, respond with the relative path to the source file of the SUT inside <sut> tags.
+Once you are finished using tool functions, respond with the relative path to the source file of the SUT inside <sut> tags. IMPORTANT: Include the full relative path from the project root, including any directory prefixes like lib/, app/, etc.
 
 Example:
 
 If you are told to find the dependencies of `test/services/country_db_interface_test.rb`,
-then you would use the functions as explained above and ultimately respond with `<sut>./app/services/country_db_interface.rb</sut>`
+then you would use the functions as explained above and ultimately respond with `<sut>app/services/country_db_interface.rb</sut>`
+
+If the file is found at `lib/roast/workflow/workflow_initializer.rb`, respond with `<sut>lib/roast/workflow/workflow_initializer.rb</sut>` (include the lib/ prefix)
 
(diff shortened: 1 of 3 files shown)


obie:
Currently making sure that it doesn't break other more complex workflows like grading.

PR #139, opened 2025-06-03
Don't execute cmds that wait for stdin

nfgrep:
We were hanging in CI because these commands wait for stdin or a file, and we weren't giving them any.


nfgrep:
This was already fixed in a different way

PR #147, opened 2025-06-06
Add 'roast list' command to display workflows in current project

dersam:
## Summary

This PR adds a new CLI command `roast list` that displays all workflow.yml files in the conventional `roast/` directory of the current project.

## Changes

- Added a new `list` command to the CLI that:
  - Searches for all `workflow.yml` files in the `roast/` directory
  - Displays them in a user-friendly format with clear indication of workflow location
  - Shows helpful error messages when no roast directory or workflows are found
  - Provides instructions on how to run the workflows

- Added comprehensive tests covering:
  - No roast directory exists
  - Empty roast directory (no workflow files)  
  - Multiple workflows including nested ones and root workflows

## Testing

Run the new command with:
```bash
bin/roast list
```

## Example Output

When workflows are found:
```
Available workflows:

  nested/workflow3 (from project)
  . (from project)
  workflow1 (from project)
  workflow2 (from project)

Run a workflow with: roast execute <workflow_name>
```

When no roast directory exists:
```
Error: No roast/ directory found in current path
```

When roast directory is empty:
```
Error: No workflow.yml files found in roast/ directory
```

### DIFF
--- a/lib/roast.rb
+++ b/lib/roast.rb
@@ -63,6 +63,32 @@ module Roast
       end
     end
 
+    desc "list", "List workflows visible to Roast and their source"
+    def list
+      roast_dir = File.join(Dir.pwd, "roast")
+
+      unless File.directory?(roast_dir)
+        raise Thor::Error, "No roast/ directory found in current path"
+      end
+
+      workflow_files = Dir.glob(File.join(roast_dir, "**/workflow.yml")).sort
+
+      if workflow_files.empty?
+        raise Thor::Error, "No workflow.yml files found in roast/ directory"
+      end
+
+      puts "Available workflows:"
+      puts
+
+      workflow_files.each do |file|
+        workflow_name = File.dirname(file.sub("#{roast_dir}/", ""))
+        puts "  #{workflow_name} (from project)"
+      end
+
+      puts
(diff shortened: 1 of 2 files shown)


obie:
## Latest Updates

I've addressed the code duplication issue and made additional architectural improvements:

### Fixed Duplication
- Removed the duplicated "spaces in step name" logic in `AgentStep#call`
- Now properly uses `name.plain_text?` from the `StepName` value object

### Architectural Improvements
1. **Proper Value Object Usage**: `BaseStep` now stores `@name` as a `StepName` object throughout the system, rather than converting between strings and `StepName` objects

2. **Duck Typing Implementation**: 
   - Added string equality to `StepName` so `step_name == "string"` works
   - Added `to_str` for implicit string conversion
   - This allows `StepName` objects to be used transparently wherever strings are expected

3. **Clean Factory Pattern**: `StepFactory` now passes `StepName` objects directly without unnecessary string conversions

These changes follow Ruby's duck typing philosophy while maintaining clean separation of concerns. The `StepName` value object provides its specialized behavior (like `plain_text?` detection) while still acting like a string in contexts that expect one.

All tests pass and the code is much cleaner\! 🎉

PR #142, opened 2025-06-05
Release v0.3.0: Upgrade to Raix 1.0

obie:
## Summary

This PR upgrades Roast to use Raix 1.0 and removes the deprecated `loop` parameter, preparing for the v0.3.0 release.

### Key Changes

- ✅ Upgraded Raix dependency from ~> 0.9.2 to ~> 1.0
- ✅ Removed deprecated `loop` parameter from all chat_completion calls (eliminates Raix deprecation warnings)
- ✅ Removed `auto_loop` attribute and all references from the codebase
- ✅ Updated all tests to work with Raix 1.0's string-only response format
- ✅ Enhanced coercion logic:
  - Boolean: empty strings now coerce to `false`
  - Iterable: properly handles JSON array strings
- ✅ Updated examples to remove `loop:` and `auto_loop:` configurations
- ✅ Bumped version to 0.3.0 with comprehensive changelog

### Breaking Changes

1. **Removed `loop` and `auto_loop` configuration options**
   - These YAML options are no longer recognized
   - Looping behavior is now automatic based on tool presence
   - Migration: Simply remove any `loop:` or `auto_loop:` settings from workflow files

2. **Raix 1.0 response format**
   - All chat completions now return strings
   - JSON responses are automatically parsed when `json: true` is specified
   - Arrays and complex response structures are no longer returned directly

### Testing

All 678 tests are passing with the new implementation.

## Test plan

- [x] Run full test suite with `bundle exec rake`
- [x] Verify no deprecation warnings from Raix
- [x] Test grading workflow to ensure no warnings in production use
- [x] Review changelog and version bump
- [x] Merge and release v0.3.0

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #148, opened 2025-06-08
Fix JSON parsing in BaseStep when json option is true

geshafer:
This PR fixes a regression introduced in https://github.com/Shopify/roast/pull/142 (5302b0b) where JSON responses were being double-parsed, causing errors for steps that use the `json: true` option. The fix removes unnecessary JSON parsing in `BaseStep#chat_completion` since [Raix 1.0 already returns parsed JSON objects when the json option is enabled](https://github.com/OlympiaAI/raix/blob/7d290d6109689eae16cda74795bba20471806596/lib/raix/chat_completion.rb#L242).

### Background

In the Raix 1.0 upgrade , the `BaseStep#chat_completion` method was modified to parse JSON responses. However, this approach incorrectly assumed that `chat_completion` always returns a string. When examining the Raix gem source code (v1.0.0), we can see that the `chat_completion` method already handles JSON parsing when `json: true` is passed. This means that when `json: true` is specified, Raix returns a parsed Ruby Hash/Array, not a JSON string. Attempting to parse it again with `JSON.parse` results in errors.

### Changes Made

1. **Fixed BaseStep#chat_completion** - Removed the redundant JSON parsing logic
2. **Updated test expectations** - Modified all tests to expect parsed JSON objects (Hash/Array) instead of JSON strings when `json: true`:
   - `test/roast/workflow/base_step_json_test.rb` - Updated all JSON response stubs to return parsed objects
   - `test/roast/workflow/prompt_step_print_response_test.rb` - Modified mock workflow to return parsed JSON
   - `test/roast/workflow/inline_prompt_configuration_test.rb` - Updated test expectation for JSON responses

### Impact

This fix ensures that workflows using JSON steps work correctly with Raix 1.0. Without this fix, any step configuration with `json: true` would fail with parsing errors.

### DIFF
--- a/lib/roast/workflow/base_step.rb
+++ b/lib/roast/workflow/base_step.rb
@@ -45,19 +45,10 @@ module Roast
         json = @json if json.nil?
         params = @params if params.nil?
 
-        workflow.chat_completion(openai: workflow.openai? && model, model: model, json:, params:).tap do |result|
-          process_output(result, print_response:)
+        result = workflow.chat_completion(openai: workflow.openai? && model, model: model, json:, params:)
+        process_output(result, print_response:)
 
-          begin
-            if json
-              return nil if result.strip.empty? # Explicitly handle empty string
-
-              return JSON.parse(result)
-            end
-          rescue JSON::ParserError
-            # If JSON parsing fails, leave it as a string
-          end
-        end
+        result
       end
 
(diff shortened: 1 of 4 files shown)


obie (approved):
Thanks for fixing my boneheaded regression!

PR #257, opened 2025-06-12
Fix model fallback for workflows without explicit model specification

jonallured:
## Summary

This PR fixes a regression where workflows without an explicit model specification would fail with an "invalid model ID" error. The fix ensures that workflows properly fall back to a default model when no model is specified.

## Problem

When a workflow YAML file doesn't include a `model:` field, the workflow execution fails with:
```
Chat completion failed\!\!\!\!\!\!\!\!\!\!\!\!\!\!\!\!: {"error" => {"message" => "invalid model ID", "type" => "invalid_request_error"}}
```

This is a regression because workflows should be able to run without explicitly specifying a model, falling back to a sensible default.

## Solution

1. **Set fallback model on workflow instance**: Modified `WorkflowRunner#create_workflow` to set the workflow's model to either the configuration's model or `StepLoader::DEFAULT_MODEL`
   
2. **Fix model format**: Corrected the `DEFAULT_MODEL` constant from `"openai/gpt-4o-mini"` to `"gpt-4o-mini"` to match the expected format

3. **Update tests**: Fixed two tests that were expecting the old model format

## Testing

- Added a new example (`examples/no_model_fallback/`) that demonstrates a workflow without model specification
- Verified the example fails before the fix and succeeds after
- All existing tests pass with the changes
- No linting issues

## Example Usage

Before this fix, this workflow would fail:
```yaml
name: my_workflow
tools:
  - Roast::Tools::ReadFile
steps:
  - analyze_file
  - process_results
```

After this fix, it automatically uses the default model (`gpt-4o-mini`).

## Breaking Changes

None. This change maintains backward compatibility while fixing the regression.

### DIFF
--- a/examples/no_model_fallback/README.md
+++ b/examples/no_model_fallback/README.md
@@ -1,0 +1,17 @@
+# No Model Fallback Example
+
+This example demonstrates the issue where workflows without explicit model specification do not properly fall back to a default model.
+
+## Purpose
+
+This workflow is based on the interpolation example but intentionally omits the `model` field to test the fallback behavior.
+
+## Expected Behavior
+
+The workflow should fall back to a default model when no model is specified at the workflow level.
+
+## Usage
+
+```bash
+bin/roast examples/no_model_fallback/workflow.yml --file examples/no_model_fallback/sample.rb
+```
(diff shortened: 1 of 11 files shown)


obie (approved):
thank you

PR #308, opened 2025-07-02
Add option for coding agent steps to resume specific session

juniper-shopify:
# Add option for coding agent steps to resume specific session

This PR adds a `resume` configuration option for coding agent steps, allowing workflows to resume Claude Code sessions from specific previous steps rather than just the immediately preceding one.

## What's new
- Added `resume` attribute to AgentStep that accepts a step name
- CodingAgent now tracks and stores session IDs in workflow output
- Session IDs are automatically extracted from Claude Code's JSON responses
- The `--resume <session_id>` flag is used when resuming from a previous step

## Example usage
```yaml
steps:
  - ^analyze the codebase
  - ^implement the feature
  - ^do something one-off # does not continue the previous session
  - ^fix the tests

implement the feature:
  continue: true  # continues from previous step

fix the tests:
  resume: implement the feature  # resumes session from specific earlier step
```


juniper-shopify:
## Question for Reviewers

@dersam 

Session IDs are stored in the output hash. If the output of the step is just a string, it needs to be converted into a hash so the session ID can be added. Example:

```
{
  "output": {
    "step_name": "some string output"
  }
}
```

becomes

```
{
  "output": {
    "step_name": {
      "content": "some string output",
      "session_id": "..."
    }
  }
}
```

Does this seem like an acceptable solution? I could also add a separate container in the workflow parallel to `output` to hold the session ids.

dersam:
Let's keep output confined to results from running a step. The session id is step metadata, not output - I'd be concerned about polluting the output with agent invocation flags.

I do think it should be possible to access this data within a step, but it should not be part of the default output hash - you should have to call something else to read it.

PR #312, opened 2025-07-08
Add script shell step

juniper-shopify:
## Summary

Add support for shell script steps in Roast workflows, allowing users to execute shell scripts directly as workflow steps alongside Ruby steps and AI prompts.

## What's New

This PR introduces a new `ShellScriptStep` class that enables running shell scripts as first-class workflow steps. The feature includes:

### Core Functionality
- **Shell Script Execution**: Execute `.sh` files directly as workflow steps
- **Environment Integration**: Automatically pass workflow context to scripts via environment variables:
  - `ROAST_WORKFLOW_RESOURCE` - Current workflow resource
  - `ROAST_STEP_NAME` - Current step name  
  - `ROAST_WORKFLOW_OUTPUT` - Previous step outputs as JSON
- **Custom Environment Variables**: Support for step-specific environment variables
- **Error Handling**: Configurable error handling with `exit_on_error` option

### Output Processing
- **Text Output**: Return raw stdout by default
- **JSON Output**: Parse stdout as JSON when `json: true` is configured
- **Error Output**: Return stderr when `exit_on_error: false` and script fails

### Discovery and Loading
- **Automatic Discovery**: Shell scripts are discovered alongside Ruby steps in:
  - Per-step paths (configured via `path:` option)
  - Phase-specific directories (e.g., `setup/`, `teardown/`)
  - Context path (workflow directory)
  - Shared directory (`../shared/`)
- **File Extension**: Use `.sh` extension for shell script steps
- **Executable Validation**: Ensures scripts are executable before running

### Step Configuration
Shell script steps support the same configuration options as other steps:
```yaml
my_script:
  json: true              # Parse output as JSON
  exit_on_error: false    # Don't fail workflow on non-zero exit
  env:                    # Custom environment variables
    CUSTOM_VAR: "value"
```

### Basic Shell Script Step
```yaml
# workflow.yml
steps:
  - setup_environment
  - run_tests
  - cleanup

# setup_environment.sh
#!/bin/bash
export DATABASE_URL="sqlite://test.db"
echo "Environment configured"
```

### JSON Output Processing
```yaml
# workflow.yml
analyze_metrics:
  json: true

# analyze_metrics.sh  
#!/bin/bash
echo '{"cpu_usage": 45.2, "memory_usage": 78.1}'
```

### Error Handling
```yaml
# workflow.yml
optional_task:
  exit_on_error: false

# optional_task.sh
#!/bin/bash
# This script may fail but won't stop the workflow
some_command_that_might_fail
```

## Implementation Details

- **Base Class**: Extends `BaseStep` for consistency with existing step types
- **Validation**: Checks file existence and executable permissions before execution
- **Security**: Uses `Open3.capture3` for safe subprocess execution
- **Environment**: Merges workflow context with custom environment variables
- **Output Processing**: Handles both text and JSON output formats
- **Error Handling**: Supports both fail-fast and continue-on-error modes

## Testing

Comprehensive test coverage includes:
- ✅ Successful script execution
- ✅ Error handling with both `exit_on_error` modes
- ✅ Environment variable passing
- ✅ JSON output parsing
- ✅ File validation (existence, permissions)
- ✅ Custom environment variables

## Files Changed

- `lib/roast/workflow/shell_script_step.rb` - New shell script step implementation
- `lib/roast/workflow/step_loader.rb` - Updated to discover and load shell scripts
- `lib/roast.rb` - Added `shellwords` require for proper shell escaping
- `test/workflow/shell_script_step_test.rb` - Comprehensive test suite


ericproulx (approved):

lib/roast/workflow/shell_script_step.rb
```diff
@@ -0,0 +1,112 @@
...
+        env_vars = {}
+
+        # Add workflow context as environment variables
+        env_vars["ROAST_WORKFLOW_RESOURCE"] = workflow.resource.to_s if workflow.resource
+        env_vars["ROAST_STEP_NAME"] = name.value
+
+        # Add workflow outputs as JSON
+        if workflow.output && !workflow.output.empty?
+          env_vars["ROAST_WORKFLOW_OUTPUT"] = JSON.generate(workflow.output)
+        end
```
geshafer, line 64: 👍 As I reflected on it, this was really the power of supporting a built in step over just using inline bash with `$()`. The whole nature of your shell script can change based on the input from a workflow.

lib/roast/workflow/step_loader.rb
```diff
@@ -60,10 +60,15 @@ def load(step_name, step_key: nil, is_last_step: nil, **options)
...
-          return load_ruby_step(step_file_path, name.to_s, is_last_step:)
+        # Look for Ruby or shell script file in various locations
+        step_file_info = find_step_file(name.to_s, per_step_path)
+        if step_file_info
+          case step_file_info[:type]
+          when :ruby
+            return load_ruby_step(step_file_info[:path], name.to_s, is_last_step:)
+          when :shell
+            return load_shell_script_step(step_file_info[:path], name.to_s, step_key, is_last_step:)
+          end
```
geshafer, line 71: Could we add shell step tests to [the step loader tests](https://github.com/Shopify/roast/blob/main/test/roast/workflow/step_loader_test.rb)?

lib/roast/workflow/shell_script_step.rb
geshafer: Do we think it's best to rely on environment variables or should we fallback to the unix narrow waist of stdin?

geshafer (changes requested):
Overall, looks good! Only blockers are step loader tests and a quick update to the readme.

I do wonder about the trade-offs of just using inline bash steps to point at local scripts. It seems likely that if someone has a tool they want a workflow to use they will likely want to use it themselves outside of that workflow. That said, I really really like the idea of keeping these scripts with the rest of the workflow logic. It also makes me think about how we could add more support for other step types.

lib/roast/workflow/shell_script_step.rb
juniper-shopify: I think trying to shove everything into stdin will become cumbersome, especially if/when there's more context to provide to the script than just the workflow output.

Doing it via environment variables also let's us deprecate old formats and iterate on them. We could, for instance, pass the workflow output in in a different format via `env_vars["ROAST_WORKFLOW_OUTPUT_V2"]` and deprecate `ROAST_WORKFLOW_OUTPUT` to be removed in the next major version. It'd be a lot harder to do that if everything is going through stdin.

lib/roast/workflow/step_loader.rb
juniper-shopify: Done! thanks for calling that out.

geshafer (approved):
Looks good! Thanks!

PR #315, opened 2025-07-10
Add verbose to configuration, and use that to control output in more places

technicalpickles:
I was looking at the output, and noticed that some of them aren't super useful for when you are just running a flow. I applied that to outputting the initializers, and I'm thinking about what other places we could use it.


dersam (changes requested):
Good idea. Add some tests and I'll merge.

PR #316, opened 2025-07-10
Add emoji in a several spots so they stand out more

technicalpickles:
I found it kinda hard to easily scan the output of a workflow. We already use emoji in a few spots, so I figure, why not more?

I also removed the 'starting workflow' message, since it's kind of implied.


technicalpickles:
Before:
CleanShot 2025-07-10 at 16 41 35@2x

After
CleanShot 2025-07-10 at 16 45 42@2x

PR #320, opened 2025-07-13
Add metadata store for tracking internal execution details

juniper-shopify:
# Add metadata store for tracking internal execution details

## What changed

- Added `MetadataManager` class to manage workflow metadata separately from user-visible output
- Integrated metadata tracking into `BaseWorkflow` with delegation to the metadata manager
- Created `MetadataAccess` helper module for tools to access metadata via Thread locals
- Updated `StepExecutorCoordinator` to set Thread locals for current step name and workflow metadata
- Extended `StateManager` to persist metadata alongside other workflow state
- Enhanced `ReplayHandler` to restore metadata when replaying workflows
- Added comprehensive test coverage for all new metadata functionality
- Updated many tests to include mock metadata along with their other fixtures
- Additional integration tests for parallel execution and interpolation will be added in a follow-up PR: #321 

## Why

This adds a parallel data structure to workflow output specifically for internal metadata that shouldn't be user-facing.
Tools and steps can now track internal state that should be programmatically available to them between invocations
without polluting the user-visible output. The metadata is automatically persisted and restored during workflow replay,
ensuring consistent behavior across sessions.

## Future considerations

The current implementation uses Thread locals to provide metadata access to tools. This approach was chosen for its simplicity and backward compatibility:

**Pros of Thread locals:**
- Tools can access metadata without changes to their method signatures
- No breaking changes to existing tool interfaces
- Simple implementation that works with the current architecture
- Automatic cleanup when threads complete

**Cons of Thread locals:**
- Implicit state that can be harder to test and debug
- Potential issues in multi-threaded scenarios if not carefully managed
- Less explicit data flow makes the code harder to understand
- Risk of accessing stale data if Thread locals aren't properly cleared

**Alternative architecture for future consideration:**
A more explicit approach would be to pass a context object to all tools that includes both the workflow and metadata references. This would require:
- Updating all tool interfaces to accept a context parameter
- Migrating existing tools to use the new interface
- Potentially supporting both patterns during a transition period

This would make data flow more explicit and testable, but requires more significant changes to the codebase.

### DIFF
--- a/lib/roast/workflow/metadata_manager.rb
+++ b/lib/roast/workflow/metadata_manager.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Roast
+  module Workflow
+    # Manages workflow metadata, providing a structure parallel to output
+    # but specifically for internal metadata that shouldn't be user-facing
+    class MetadataManager
+      def initialize
+        @metadata = ActiveSupport::HashWithIndifferentAccess.new
+        @metadata_wrapper = nil
+      end
+
+      # Get metadata wrapped in DotAccessHash for dot notation access
+      def metadata
+        @metadata_wrapper ||= DotAccessHash.new(@metadata)
+      end
+
+      # Set metadata, ensuring it's always a HashWithIndifferentAccess
+      def metadata=(value)
+        @metadata = if value.is_a?(ActiveSupport::HashWithIndifferentAccess)
+          value
+        else
+          ActiveSupport::HashWithIndifferentAccess.new(value)
+        end
+        # Reset the wrapper when metadata changes
+        @metadata_wrapper = nil
+      end
+
+      # Get the raw metadata hash (for internal use)
+      def raw_metadata
+        @metadata
+      end
+
+      # Get a snapshot of the current state for persistence
+      def to_h
+        @metadata.to_h
+      end
+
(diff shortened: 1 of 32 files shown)


ericproulx (approved):

lib/roast/workflow/metadata_manager.rb
```diff
@@ -0,0 +1,47 @@
...
+    # but specifically for internal metadata that shouldn't be user-facing
+    class MetadataManager
+      def initialize
+        @metadata = ActiveSupport::HashWithIndifferentAccess.new
+        @metadata_wrapper = nil
+      end
+
+      # Get metadata wrapped in DotAccessHash for dot notation access
+      def metadata
+        @metadata_wrapper ||= DotAccessHash.new(@metadata)
```
ericproulx, line 15: DotAccessHash is really convenient but it adds another to access properties and I don't know if we need it for consistency. For instance, this [PR](https://github.com/Shopify/roast/blob/2bca1cb73696a64295dc4a6a6ac32d61a657e854/lib/roast/workflow/agent_step.rb#L27) does not use it and a IMO, few people would instinctively use the dot notation. Don't get me wrong, I like the dot notation but for consistency, its adding a new "standard"   in the codebase. What do you think ?

PR #321, opened 2025-07-13
Add additional metadata integration tests

juniper-shopify:
# Add additional integration tests for metadata functionality

## What changed

- Added integration tests for metadata behaviour during parallel step execution
- Added test for metadata interpolation in workflow expressions using `{{metadata.path.to.value}}` syntax

## Why

These tests verify that the metadata system works correctly in more complex scenarios:

1. **Parallel execution tests** verify that:
   - Metadata access is thread-safe when steps run in parallel
   - Each parallel step can write its own metadata without corruption
   - Sequential steps before and after parallel execution can access all metadata
   - Thread locals are properly isolated between parallel executions

2. **Interpolation test** verifies that:
   - Metadata can be interpolated in step prompts using the same `{{ }}` syntax as output
   - Both dot notation (`metadata.step.field`) and bracket notation (`metadata['step']['field']`) work
   - The interpolation integrates properly with the existing interpolator infrastructure

## Test details

The parallel execution tests demonstrate that the metadata system maintains data integrity even when multiple steps execute concurrently, which is critical for workflow reliability.

The interpolation test ensures metadata is a first-class citizen in the workflow system, accessible through the same expression syntax users are already familiar with for output values.

## Related PRs

This PR adds integration tests for the metadata functionality introduced in #320. These tests were separated to keep the initial implementation PR focused and easier to review.

### DIFF
--- a/test/roast/workflow/workflow_executor_test.rb
+++ b/test/roast/workflow/workflow_executor_test.rb
@@ -314,4 +314,39 @@ class RoastWorkflowWorkflowExecutorTest < ActiveSupport::TestCase
     result = @executor.interpolate("Analysis result: {{output['dangerous_output']}}")
     assert_equal "Analysis result: #{dangerous_content}", result
   end
+
+  test "interpolates metadata from workflow metadata store" do
+    Dir.mktmpdir do |tmpdir|
+      # Create a real workflow with a real metadata manager
+      workflow_file = File.join(tmpdir, "test_workflow.yml")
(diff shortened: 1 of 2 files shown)


test/roast/workflow/workflow_executor_test.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 require "test_helper"
+require "tmpdir"
```
ericproulx, line 4: "tmpdir" is already required in `test_helper`

ericproulx (approved):

PR #326, opened 2025-07-15
Make timeouts more robust

nfgrep:
- Don't set a MAX_TIMEOUT, we shouldn't assume what type of tasks folks are using the tool for.
- Offer a GRACE_PERIOD between `TERM` and `KILL` to give processes a chance to cleanup children.
- Invoke processes to be timed out with their own pgroup, so we can reliably kill all the child processes regardless of the parent's behaviour.

### DIFF
--- a/lib/roast/helpers/timeout_handler.rb
+++ b/lib/roast/helpers/timeout_handler.rb
@@ -18,6 +18,9 @@ module Roast
     #
     # @example With custom grace period
     #   output, status = TimeoutHandler.call("sleep 10", timeout: 10, grace_period: 1)
+    #
+    # @example Disable process group for TTY commands (like macOS security)
+    #   output, status = TimeoutHandler.call("security find-generic-password", timeout: 30, use_pgroup: false)
     class TimeoutHandler
       DEFAULT_TIMEOUT = 30
       DEFAULT_GRACE_PERIOD = 5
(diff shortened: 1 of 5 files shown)


lib/roast/helpers/timeout_handler.rb
```diff
@@ -15,26 +15,35 @@ module Helpers
...
           output = ""
           exit_status = nil
           wait_thr = nil
 
           begin
             Timeout.timeout(timeout) do
-              stdin, stdout, stderr, wait_thr = Open3.popen3(command, chdir: working_directory)
+              popen3_options = { chdir: working_directory }
+              popen3_options[:pgroup] = true if use_pgroup
+              stdin, stdout, stderr, wait_thr = Open3.popen3(command, popen3_options)
```
ericproulx, line 46: Since you're in this area, we could change `popen3` by `popen2e` which merges stdout and stderr. Line 48 could be removed

lib/roast/helpers/timeout_handler.rb
```diff
@@ -15,26 +15,35 @@ module Helpers
...
         # @raise [Timeout::Error] When command exceeds timeout duration
-        def call(command, timeout: DEFAULT_TIMEOUT, working_directory: Dir.pwd)
+        def call(command, timeout: DEFAULT_TIMEOUT, working_directory: Dir.pwd, grace_period: DEFAULT_GRACE_PERIOD, use_pgroup: true)
           timeout = validate_timeout(timeout)
           output = ""
           exit_status = nil
           wait_thr = nil
 
           begin
             Timeout.timeout(timeout) do
```
ericproulx, line 43: I found a [gist](https://gist.github.com/lpar/1032297) that replaces the `Timeout` by something cleaner. Seems worth trying :)

Issue #297, opened 2025-06-24
Noisy output when running tests :(


ericproulx:
When running tests, there's a lot of noisy output.  This is caused by the usage of `$stderr.puts` inside classes.

There are plenty of ways to fix that:

- Surround function calls that uses `$stderr.puts` with a `capture_io`. The last comes with `minitest`
- Create a dependency injections that we could manage `$stderr.puts`
- Stubbing `$stderr` for tests
- ...

I like the dependency injection. Any other suggestions ?

Issue #120, opened 2025-05-31
Many  method redefined warnings during tests from raix


technicalpickles:
I'm seeing a ton of warnings when running `rake test`:

```
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old search_for_file
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: previous definition of search_for_file was here
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old read_file
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: previous definition of read_file was here
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old grep
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: previous definition of grep was here
./Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old search_for_file
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: previous definition of search_for_file was here
/Users/josh.nichols/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/raix-0.8.6/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old read_file
[...]

PR #338, opened 2025-07-30
Add ripgrep check to grep tool

TravisTheTechie:
Tests were failing because ripgrep was not installed. This adds a hint for the user to install it.

```
% brew remove rg
Uninstalling /opt/homebrew/Cellar/ripgrep/14.1.1... (14 files, 6MB)
% bundle exec rake
....
Finished in 30.431200s, 32.3681 runs/s, 83.1055 assertions/s.

  1) Failure:
RoastToolsGrepTest#test_limits_output_to_MAX_RESULT_LINES [test/roast/tools/grep_test.rb:66]:
Expected /truncated to 100 lines/ to match "Error grepping for string: ripgrep is not available. Please install it using your package manager (e.g., brew install rg) and make sure it's on your PATH.".

  2) Failure:
RoastToolsGrepTest#test_handles_curly_braces_in_search_string [test/roast/tools/grep_test.rb:53]:
Expected /react_file\.js/ to match "Error grepping for string: ripgrep is not available. Please install it using your package manager (e.g., brew install rg) and make sure it's on your PATH.".

  3) Failure:
RoastToolsGrepTest#test_searches_for_string_using_ripgrep [test/roast/tools/grep_test.rb:33]:
Expected /test_file1\.txt/ to match "Error grepping for string: ripgrep is not available. Please install it using your package manager (e.g., brew install rg) and make sure it's on your PATH.".

  4) Failure:
RoastToolsGrepTest#test_handles_errors_gracefully [test/roast/tools/grep_test.rb:42]:
--- expected
+++ actual
@@ -1 +1 @@
-"Error grepping for string: Command failed"
+"Error grepping for string: ripgrep is not available. Please install it using your package manager (e.g., brew install rg) and make sure it's on your PATH."


985 runs, 2529 assertions, 4 failures, 0 errors, 5 skips
...
% brew install rg
...
% bundle exec rake
...
Finished in 30.656391s, 32.1303 runs/s, 82.6908 assertions/s.

985 runs, 2535 assertions, 0 failures, 0 errors, 5 skips
...
%
```

PR #341, opened 2025-07-31
Test warning clean up

TravisTheTechie:
1. Multiple `include` calls in `workflow_initializer.rb`; e.g.,
```
/Users/travis/.gem/ruby/ruby-3.4.2/gems/raix-openai-eight-1.0.1/lib/raix/function_dispatch.rb:67: warning: method redefined; discarding old search_for_file
/Users/travis/.gem/ruby/ruby-3.4.2/gems/raix-openai-eight-1.0.1/lib/raix/function_dispatch.rb:67: warning: previous definition of search_for_file was here
```

Called out in #120. 

2. `workflow_initializer_test.rb` was not resetting the `BaseWorkflow` constant between tests.

This was a bit complicated, but the tests ended up coupled between execution. This removes that coupling.
Also changed the assertions from what APIs were called to what is the behavior of the targeted object.
(Behaviors vs. implementation.)

3. `warning: redefining 'object_id' may cause serious problems` warning fixed. Had no impact on the test, so removed.

Doesn't really address #297, but makes some progress towards it.

### DIFF
--- a/lib/roast/tools/grep.rb
+++ b/lib/roast/tools/grep.rb
@@ -27,6 +27,10 @@ module Roast
       def call(string)
         Roast::Helpers::Logger.info("🔍 Grepping for string: #{string}\n")
 
+        unless system("command -v rg >/dev/null 2>&1")
+          raise "ripgrep is not available. Please install it using your package manager (e.g., brew install rg) and make sure it's on your PATH."
+        end
+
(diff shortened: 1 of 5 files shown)


TravisTheTechie:
This branch is cut off #338. If we can't merge #338 for some reason, I'll rebase this off main.

PR #359, opened 2025-08-16
Initial setup of Sorbet

dersam:
Adding static typechecking will make it easier to navigate and refactor the codebase. Initial setup already revealed two invalid constants that were undefined and never reached.

I will gradually enable typing on other files in followup PRs. I have intentionally not included runtime typechecking as I don't think the performance cost is worth the benefit.

This initial PR sets up Sorbet typechecking with `bin/srb tc`, but does not add any signatures yet.


nfgrep:
> I will gradually enable typing on other files in followup PRs. I have intentionally not included runtime typechecking as I don't think the performance cost is worth the benefit.

Agreed 👍

nfgrep (approved):

PR #366, opened 2025-08-18
Define Ruby DSL executor

dersam:
Add basic executor for a Ruby DSL. Further steps will be defined in the executor (such as agent definitions and step type inference as with the current yml standard).

This basic version lets us start iterating on the DSL design ([notes from earlier whiteboarding](https://github.com/Shopify/roast/blob/incubate-ruby-dsl/dsl.md?plain=1) with @juniper-shopify and @nfgrep). Nobody should use this yet, which is why it's hidden behind an explicit flag.

With `--executor=dsl`
Screenshot 2025-08-19 at 10 22 12 AM

### DIFF
--- a/dsl/simple.rb
+++ b/dsl/simple.rb
@@ -1,0 +1,10 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Executor
+
+# This is a dead simple workflow that calls two shell scripts
+shell <<~SHELLSTEP
+  echo "I have no idea what's going on"
+SHELLSTEP
+shell "pwd"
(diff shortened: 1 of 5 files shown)


dblock:
Maybe this can be more generic and then there's no need for a `--experimental_ruby_dsl` flag? For example, this could be an optional `--executor=dsl` (`--executor=default` by default) option. Then I'd document this in README and in console output with a "warning: experimental".

dersam:
> Maybe this can be more generic and then there's no need for a `--experimental_ruby_dsl` flag? For example, this could be an optional `--executor=dsl` (`--executor=default` by default) option. Then I'd document this in README and in console output with a "warning: experimental".

Sure, makes sense to me.

dsl/simple.rb
nfgrep, line 1: For later but eventually `examples/dsl` or smth

nfgrep (approved):

dsl/simple.rb
dersam: I left it out of examples to avoid temptation to use the experimental feature. We can move it there once it's more robust.

PR #386, opened 2025-08-25
Copy example workflows in test sandboxes

dersam:
Adds an easy way to import workflow examples from the project `examples/` directory when using the `in_sandbox` test function. If the symbol passed to `with_workflow` matches a directory in `examples/`, the entire example will be copied into the `roast` directory in the sandbox.

This allows us to validate our examples via testing. If a feature would break an example, tests will fail.

Additionally, testing the examples by definition is testing features we care about and expose to users, in the way we expect them to be used.

Examples can be tested as follows:

```ruby
  test "example LLM workflow" do
    in_sandbox with_workflow: :apply_diff_demo do
      roast("execute", "apply_diff_demo")
    end
  end
```

### DIFF
--- a/test/support/functional_test.rb
+++ b/test/support/functional_test.rb
@@ -10,16 +10,23 @@ class FunctionalTest < ActiveSupport::TestCase
   # with_workflow will create the workflow defined in Workflows.:name
   # Returns an array of strings [stdio_output, stderr_output]
   def in_sandbox(with_workflow: nil)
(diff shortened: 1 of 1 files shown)


juniper-shopify (approved):

dersam:
### Merge activity

* **Aug 27, 3:38 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/386).

PR #387, opened 2025-08-26
Example LLM workflow tests

dersam:
Add tests for example workflows using VCR. 

The LLM interaction is recorded for real, then it can be replayed deterministically on all subsequent test runs by reusing the recorded http request/response.


test/support/functional_test.rb
```diff
@@ -11,12 +11,25 @@ def roast(*args)
...
+    root_project_path = Dir.pwd
+    project_dot_roast_path = File.join(root_project_path, ".roast")
+    candidate_example_path = File.join(root_project_path, "examples", with_workflow.to_s)
 
     capture_io do
       Dir.mktmpdir do |dir|
         Dir.chdir(dir) do |dir|
-          Dir.mkdir(".roast")
+          if Dir.exist?(project_dot_roast_path) && ENV["RECORD_VCR"]
+            FileUtils.cp_r(project_dot_roast_path, ".roast")
```
dersam, line 22: This loads your local roast initializers when running the test, so we can record the HTTP interactions where necessary.

test/support/functional_test.rb
```diff
@@ -11,12 +11,25 @@ def roast(*args)
...
+          else
+            Dir.mkdir(".roast")
+            Raix.configure do |config|
+              config.openai_client = OpenAI::Client.new(
```
dersam, line 26: We have to define a client or Raix will crash.

juniper-shopify: Raix is very annoying to configure. You can't just add the roast-ai gem to a project and set up a workflow directory and have it work :/

test/fixtures/vcr_cassettes/available_tools.yml
```diff
@@ -0,0 +1,1027 @@
+---
```
juniper-shopify, line 1: These are super verbose, which kind of threw me at first, but tbh I like that it is very thorough and explicit.

test/functional/roast_examples_test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require "test_helper"
+
+# This test suite validates the example workflows in `examples/`.
+class RoastExamplesTest < FunctionalTest
+  test "available_tools_demo workflow" do
+    VCR.use_cassette("available_tools") do
+      out, _ = in_sandbox with_workflow: :available_tools_demo do
+        roast("execute", "available_tools_demo")
+      end
```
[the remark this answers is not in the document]
juniper-shopify: +1

I don't think it makes sense to invest too much more into the existing examples using the legacy config syntax, at this point, though. We can keep some of them, I'm sure, and ideally set it up to test that they actually work, but ones that don't/can't actually run in tests we should at least consider removing.

It would be great to set up a test_helper framework from the beginning though so that examples we write using the new DSL can actually be run in tests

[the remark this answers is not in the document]
dersam: Agreed. Part of what I'll be doing with these tests is auditing our existing examples. Once all examples are tested, I'll add a cop to require a test for any new example.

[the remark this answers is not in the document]
juniper-shopify: Though, immediately I consider a problem:

We want roast core to be minimal, and for coding agents and tools and other functionality to be pulled in as plugins. that makes it hard for us to ship examples that can actually be run using just the roast core gem.

It'd also, in that vein, be weird to even have those examples in the core roast repo because they'd be referencing other gems on which roast core does not depend, and shouldn't have to be aware od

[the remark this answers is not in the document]
juniper-shopify: I think we may want to end up with a roast-examples repo that depends on core roast and a bunch of the standard plugins we provide to show a comprehensive set of use cases. Food for thought. It would of course be good if integration tests in the core gem would catch changes that break our standard plugins' integration. We could have a CI step for core roast that runs integration tests in the examples project maybe. kind of a weird dependency structure, though.

Kind of just thinking out loud here. This PR looks good and I like your plan

juniper-shopify (approved):

test/fixtures/vcr_cassettes/available_tools.yml
dersam: It's generated by VCR, which automatically records real HTTP requests for reuse as test fixtures. I didn't write this by hand. :D

dersam:
### Merge activity

* **Aug 27, 3:38 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/387).

PR #394, opened 2025-08-29
Add sample workflow to analyze files

dersam:
Add a tested example workflow `basic_prompt_workflow` that reads a CSV file and answers some simple questions.

LLM responses in the test are mocked using VCR recordings of actual LLM interactions.

### DIFF
--- a/test/support/functional_test.rb
+++ b/test/support/functional_test.rb
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 
 class FunctionalTest < ActiveSupport::TestCase
+  setup do
+    Thread.current[:chat_completion_response] = nil
+    Thread.current[:current_step_name] = nil
+    Thread.current[:workflow_metadata] = nil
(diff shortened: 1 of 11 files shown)


test/support/functional_test.rb
```diff
@@ -1,6 +1,17 @@
...
 class FunctionalTest < ActiveSupport::TestCase
+  setup do
+    Thread.current[:chat_completion_response] = nil
+    Thread.current[:current_step_name] = nil
+    Thread.current[:workflow_metadata] = nil
+    Thread.current[:workflow_context] = nil
+    Thread.current[:step] = nil
+    Thread.current[:result] = nil
+    Thread.current[:error] = nil
+    Roast::Helpers::Logger.reset
```
dersam, line 12: I have no idea what the logger is doing, but it will stop outputting the results of tool calls after running one test. wtf.

PR #403, opened 2025-09-04
refactor(base_step): use workflow/steploader default model

tungnt1203:
Summary
- Remove hardcoded default "anthropic:claude-opus-4".
- Model resolution: explicit param → workflow.model → StepLoader::DEFAULT_MODEL ("gpt-4o-mini").
- Remove default model in StepLoader


tungnt1203:
@dersam  thanks for the review! I removed model fallback from StepLoader; only apply per-step model when explicitly set. tests updated accordingly. 
let me know if anything else needs attention.

PR #404, opened 2025-09-05
Add test grading workflow example

dersam:
Restore the old test grading example workflow. I've simplified it by removing the test coverage pieces, as that can vary wildly depending on the system and this is meant to illustrate how to use roast.

I've made some enhancements to functional testing. You can now dump the captured stdout and stderr to a file, and reuse this file as an automatic test fixture. `in_sandbox` will automatically attempt to assert the output against a fixture with the matching name, if present. This makes it a lot easier to define new functional tests against a convention.


obie (approved):
LGTM. Glad you readded

Issue #409, opened 2025-09-10
Interactive input step does not work as expected


ShopiRubio:
I want to be able to pass an argument that is **not** a file to my roast workflow.

The `-t` flag seems to be strictly for files as it makes it available as {{file}} in my workflow.yml. It assumes it is a file and even adds a proper path to it. Thus, I was looking for something else I could use. 

I tried using [Input Step](https://github.com/Shopify/roast#:~:text=Input%20step%3A%20Interactive%20prompts%20for%20user%20input%20during%20workflow%20execution) for the user to enter the argument:

```yml
name: Test

tools:
  - Roast::Tools::Grep

steps:
  - ask_for_table:
      prompt: "Indicate the table name."
      type: text

  - table_name: >
      $(echo "{{output.ask_for_table}}")
```

Unfortunately, it doesn't work as expected:

```console
me@shopify shopify % bundle exec roast execute roast/test/workflow.yml                                                                        2025-09-10T14:30:57
🔥🔥🔥 Everyone loves a good roast 🔥🔥🔥

Loading project initializers from /Users/me/.roast/initializers
Loading initializer: /Users/me/.roast/initializers/base_workflow_metrics.rb
Loading initializer: /Users/me/.roast/initializers/raix.rb
Loading initializer: /Users/me/.roast/initializers/workflow_coverage_step_metrics.rb
Loading initializer: /Users/me/.roast/initializers/workflow_specific_metrics.rb
Starting workflow...
Workflow: /Users/me/roast/test/workflow.yml
Options: {executor: "default"}
Running targetless workflow
Prompt file for workflow not found: 
Executing: Indicate the table name. (Resource type: none)
🔍 Grepping for string: table
✓ Complete: ask_for_table (consumed 5,629 tokens, total 5,629)


Executing: table_name (Resource type: none)
✓ Complete: table_name (consumed 0 tokens, total 5,629)


Final output saved to: test_0021e3d6_20250910_183149_342
🔥🔥🔥 Final Output: 🔥🔥🔥
The term "table" is mentioned in the context of SQL keywords, specifically in the following line:

``javascript
var sqlKeywords = "alter and as asc between by count create delete desc distinct drop from group having in insert into is join like not on or order select set table union update values where limit ";
``

This indicates that "table" is a keyword used in SQL statements, but it does not specify a particular table name. If you are looking for a specific table name in the project, please provide more context or specify what you are looking for.
🔥🔥🔥 ROAST COMPLETE! 🔥🔥🔥
[...]

PR #414, opened 2025-09-20
Fix README examples for input steps

ric2b:
The examples for input steps seem to have incorrect syntax that will call the model instead of prompting the user for input.

Closes https://github.com/Shopify/roast/issues/409


ric2b:
(I signed the CLA but can't rerun the check)

ric2b:
Noticed that the example for "choice" was also incorrect

PR #417, opened 2025-10-03
Implement cog executor

dersam:
This PR implements a block based DSL that provides a new internal Roast architecture. It prevents state pollution by design, as well as providing an easier to understand interface versus the existing YAML definitions.

Terminology: `cogs` are a step definition. It's a generic class that can be used to implement specific kinds of step types. In this PR, it's a simple CLI command.

There's a three step process to a workflow:
1. **config block**: configure cogs. Creates cogs if they don't exist, potentially sets execution-global configs on cog types (so if you want all command cogs to write their output, we can do that without having to explicitly configure each one).
2. **execute block**: Prepare the cog stack for execution. Nothing is actually run here! We preprocess the set of actions. At the end of this step, we know everything that we will possibly execute (a future export into a DTO serialized format would happen here).
3. **Execute the stack**. Map through the cogs and call `run!` on each of them. It's a bang method because you can only use an executor once. It will error out if you try to run it again in the same Roast run. This completely prevents state pollution across workflows.

Note that the DSL hides the exact implementation of the simple interface of `cmd`. Behind the scenes, `config` and `execute` are operating on completely separate objects - the meaning of the cog call is totally context dependent. This enforces a separate between configuration, definition of execution, and the actual execution, providing many more ways to validate a workflow for correctness before you run it (ultimately saving on tokens for heavy AI workflows that would error out near the end).


dersam:
* **#417** <a href="https://app.graphite.dev/github/pr/Shopify/roast/417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #393, opened 2025-08-28
Delete broken example workflows

dersam:
The provided examples don't work and will continue rotting. Instead of fixing all of these generated examples (many of which rely on missing features or hallucinated functionality), we will delete them and work on new examples that will be validated using the functional test suite.

PR #442, opened 2025-10-14
Fix #395: Add temporary messaging for broken 'New from prompt' option from roast init

### LINKED ISSUE #395: `roast init` failure
This should work in a new directory somewhere... 


```
% bundle exec roast init
🔥🔥🔥 Everyone loves a good roast 🔥🔥🔥

Select an option:
/Users/travis/.gem/ruby/ruby-3.4.2/gems/cli-ui-2.3.0/lib/cli/ui/prompt.rb:7: warning: reline was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add reline to your Gemfile or gemspec to silence this warning.
? Select initialization method: (You chose: New from prompt (beta))
Create a new workflow from a description                                                                                                                            
                                                                                                                                                                    
Error generating workflow: Workflow file not found: /Users/travis/src/github.com/Shopify/roast/examples/workflow_generator/workflow.yml
% git pull
Already up to date.
% pwd
/Users/travis/src/github.com/Shopify/roast
```

mathiusj:
## Problem

Resolves: https://github.com/Shopify/roast/issues/395

The 'New from prompt (beta)' option in roast init was broken after PR #393 deleted the workflow_generator example. The generator was removed because it relied on hallucinated Roast features and generated unreliable workflows.

## Solution
Rather than completely removing the feature, preserve user expectations with a "Coming soon" message that explains the feature is being rebuilt with proper validation and upcoming DSL integration. This maintains transparency about feature status while directing users to working examples in the meantime.

The workflow generator should be rebuilt properly with better guardrails to prevent LLM hallucination, validation that generated workflows actually work, and possible integration with upcoming DSL features.

## Testing

```bash
bin/roast init
# Select "New from prompt (coming soon)"
# Verify helpful message is shown
```

### DIFF
--- a/lib/roast.rb
+++ b/lib/roast.rb
@@ -303,7 +303,7 @@ module Roast
       end
 
       puts "Select an option:"
-      choices = ["Pick from examples", "New from prompt (beta)"]
+      choices = ["Pick from examples", "New from prompt (coming soon)"]
 
       selected = run_picker(choices, "Select initialization method:")
 
@@ -311,8 +311,8 @@ module Roast
       when "Pick from examples"
         example_choice = run_picker(examples, "Select an example:")
         copy_example(example_choice) if example_choice
-      when "New from prompt (beta)"
-        create_from_prompt
+      when "New from prompt (coming soon)"
+        show_coming_soon_message
       end
     end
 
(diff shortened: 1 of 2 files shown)


mathiusj:
* **#442** <a href="https://app.graphite.dev/github/pr/Shopify/roast/442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

juniper-shopify (changes requested):
I don't think we should indicate that we're going to keep this feature around at all. I'd prefer messaging that says the workflow framework is being rebuilt and this feature will be reconsidered and potentially redesigned if it still makes sense to do so.

dersam:
> I don't think we should indicate that we're going to keep this feature around at all. I'd prefer messaging that says the workflow framework is being rebuilt and this feature will be reconsidered and potentially redesigned if it still makes sense to do so.

Agreed, I don't want to promise a feature that we have no current plans to build.

mathiusj:
ah, I just realized my commit message was stale after pivoting to remove that feature entirely. The pr description is accurate but the message is not! I'd revert this to push the right message, but I'm not sure it's really worth it? @dersam @juniper-shopify thoughts? happy to go either way.

dersam:
Don't worry about it.

PR #444, opened 2025-10-16
Split Contexts from Managers in DSL

juniper-shopify:
# Refactor DSL implementation to separate Contexts from Managers

This PR refactors the DSL implementation to use a Manager pattern separate from the Contexts to which we bind cog methods, improving separation of concerns and code organization. Key changes include:

- Renamed `Executor` to `Workflow` to better reflect its purpose as distinct from `ExecutionContext` and `ExectionManager`
- Split context classes into context and manager classes to handle different responsibilities. Contexts are now this:
  - `ConfigContext` - the class to which the cog config methods are bound. it has no other methods. this is the context in which the workflow's `config` block is evaluated
  - `ExecutionContext` - the class to which the cog input methods are bound. it has no other methods. this is the context in which the workflows' `execute` block is evaluated
  - `CogInputContext` - the class to which the cog output methods are bound. it has no other methods. this is the context in which each cog's input block is evaluated (giving it access to the outputs of other cogs)
- Managers are now this:
  - `ConfigManager` - binds the cogs to the ConfigContext; collects and exposes the configs produced by executing the cogs' individual config blocks (in the context of the cogs' own Config classes)
  - `ExecutionManager` - binds the cogs to the ExecutionContext; creates the cog instances; runs the cog stack
  - `CogInputManager` - binds the cog output-getter methods to the CogInputContext; provides access to the output of previously run cogs when a cog's input block is evaluated
- Added a simple placeholder `Cog::Registry` to centralize cog registration
- Simplified context classes to focus on their specific roles
- Improved error handling with more specific error classes
- Added state tracking with `preparing?`, `prepared?`, `started?`, and `completed?` flags

Previously, when the manager and context concerns were combined, it was possible to write something like this:

```
execute do
  cmd(:echo) { "echo hi" }
  prepare! # this would re-invoke the `prepare!` method on the ExecutionContext/Manager, resulting in an exception
end
```


dersam (approved):

juniper-shopify:
### Merge activity

* **Oct 16, 2:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/444).
* **Oct 16, 2:54 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/444).

PR #448, opened 2025-10-16
Getters and setters for config

nfgrep:


lib/roast/dsl/cogs/chat.rb
```diff
@@ -40,40 +40,19 @@ def initialize(response)
...
-          #: (?Symbol?) -> Symbol?
-          def provider(provider_name = nil)
-            @values[:provider] ||= provider_name
-          end
-
-          #: (?bool?) -> bool?
-          def assume_model_exists(assume_model_exists = nil)
-            @values[:assume_model_exists] ||= assume_model_exists || false
-          end
+          field :model, String
```
dersam: Strong opinion: the type should not be part of the runtime definition. We can handle this with a Sorbet compiler down the line, much like how `attr_accessor` is already handled.

lib/roast/dsl/cogs/chat.rb
```diff
@@ -34,40 +34,19 @@ def initialize(response)
...
-
-          #: (?bool?) -> bool?
-          def assume_model_exists(assume_model_exists = nil)
-            @values[:assume_model_exists] ||= assume_model_exists || false
-          end
+          field :model
+          field :api_key, ENV["OPENAI_API_KEY"]
+          field :base_url, ENV["OPENAI_API_BASE_URL"]
+          field :provider, :openai
+          field :assume_model_exists, false
```
juniper-shopify: This is super elegant! I love it ❤️

PR #450, opened 2025-10-17
Add Map cog for parallel execution of workflow steps

juniper-shopify:
# Add Map Cog for Parallel Execution in Workflows

This PR introduces a new `map` cog that allows workflows to apply a scoped executor to each item in a collection. This enables more concise and readable workflow definitions when performing the same operation on multiple items.

Key changes:
- Added `SystemCogs::Map` class with appropriate input and parameter handling
- Implemented `map` cog execution in the system cog manager
- Created a params base class for system cogs to standardize parameter handling
- Added type signatures and Sorbet RBI shims for the new functionality

## New Concept - SystemCog::Params

This PR adds an additional parameterization point for system cogs. This allows the cog method bound to the `execute` context to accept additional positional and/or keyword args in addition to the cog's name.

The rationale for keeping this new functionality restricted to system cogs only is that it is largely redundant with cog input (and/or cog config) for most purposes. The specific utility of the params concept for system cogs is that it allows parameters that would otherwise be specified in the cog's input block to be determined at the time the `execute` block is evaluated, before the cog's input block is evaluated.

This will make it possible for Roast to do more substantive 'static' analysis of a workflow definition before actually running the workflow. For example, by specifying the scope of the executor that the map cog should apply to its inputs at the level of the ExecutionContext instead of the CogInputContext, Roast could determine before execution begins what a flow diagram for the workflow would look like.

### DIFF
--- a/dsl/map.rb
+++ b/dsl/map.rb
@@ -1,0 +1,45 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd do
+    print_all!
+  end
+end
+
+execute(:capitalize_a_word) do
+  cmd(:capitalize) do |my, word|
+    word ||= cmd(:word).out.strip
+    my.command = "sh"
+    my.args << "-c"
+    my.args << "echo '#{word}' | tr '[:lower:]' '[:upper:]'"
+  end
+end
+
+execute do
+  words = ["hello", "world", "goodnight", "moon"]
+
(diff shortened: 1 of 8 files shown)


dersam:
I'm trying to wrap my head around why the system cog concept is necessary to make `map` work, and I'm not getting it. (had the same question about the downstack PR and was hoping this one would make it make sense)

Can you elaborate or add a diagram?

dersam (approved):

PR #451, opened 2025-10-17
Prototyping third party gems

dersam:
Demonstration of using third party cogs and binding them into a workflow.

This adds a new syntax `use`, in which you declare the cog you want to use and the gem it is from. `use` will ensure the gem is required and then bind the cog classes into the registry.

For testing, I've set up a very simple demo gem, and a standalone "project" that contains a roast workflow with its own gemfile. This ensures that gem loading works properly in projects using Roast.

A workflow can _also_ define workflow specific cogs. If `use` does not have the `from:` parameter passed, the workflow will look in the cogs directory for that given workflow (conventionally, the `cogs/` dir in the same directory as the workflow file).

### DIFF
--- a/dsl/demo/Gemfile
+++ b/dsl/demo/Gemfile
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+gem "roast-ai", path: "../.."
+gem "plugin-gem-example", path: "../plugin-gem-example"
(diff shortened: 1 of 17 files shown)


dsl/demo/Gemfile
```diff
@@ -0,0 +1,2 @@
+gem "roast-ai", path: "../.."
+gem "plugin-gem-example", path: "../plugin-gem-example"
```
dersam: This ensures our demo is always using the local versions of gems so our examples are valid.

nfgrep (approved):

juniper-shopify (approved):

PR #462, opened 2025-10-20
Rename Execute cog to Call for better semantics

juniper-shopify:
# Rename `execute` to `call` for better clarity in DSL

This PR renames the `execute` cog to `call` to better reflect its purpose and avoid confusion with the `execute` block. The change improves the readability of workflows by making it clearer when we're calling another scope.

Changes include:
- Renamed `Cogs::Execute` class to `Cogs::Call`
- Updated all references to use `call` instead of `execute` in DSL methods
- Improved the input coercion logic in the renamed class
- Removed redundant initialization code
- Updated type signatures in Sorbet RBI files

Example usage now looks like:
```ruby
execute do
  cmd(:before) { "echo '--> before'" }
  call { :capitalize_a_random_word }
  call { :capitalize_a_random_word }
  call { :capitalize_a_random_word }
  cmd(:after) { "echo '--> after'" }
end
```

### DIFF
--- a/dsl/scoped_executors.rb
+++ b/dsl/scoped_executors.rb
@@ -21,8 +21,8 @@ end
 
 execute do
   cmd(:before) { "echo '--> before'" }
-  execute { :capitalize_a_random_word }
-  execute { :capitalize_a_random_word }
-  execute { :capitalize_a_random_word }
+  call { :capitalize_a_random_word }
+  call { :capitalize_a_random_word }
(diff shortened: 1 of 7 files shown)


juniper-shopify:
### Merge activity

* **Oct 20, 6:57 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/462).

PR #463, opened 2025-10-20
Refactor Call system cog to support named execution scopes

juniper-shopify:
# Update `call` system cog to accept scope as a parameter

This PR changes the `call` system cog to accept the executor scope as a parameter rather than through the input block.

Accepting the scope via the input block was a mistake in the initial implementation of this cog -- a design principal I consider foundational is that the shape of the workflow can be knowable in its entirety when the workflow definition is evaluated, before anything is run. In order to support this, the call relationships between blocks must be determinable without evaluating any cog's input block, which can depend on the output of running previous cogs in the workflow.

Key changes:
- Added a `Params` class to `SystemCog` to handle parameters passed directly to cog methods
- Modified `call` to accept the scope as its first parameter and an optional name as the second parameter
- Updated the execution manager to handle the new parameter structure
- Added a `generate_fallback_name` method to centralize the generation random names for unnamed cogs, since we're now invoking it from multiple places
- Simplified the cog store and cog stacks to just take a cog instance directly rather than separate name and instance

Example usage:
```ruby

cmd(:name_of_cog) {}     # Named regular cog
cmd {}                               # Anonymous regular cog

call(:capitalize_a_random_word, :name_of_call)    # Named call
call(:capitalize_a_random_word)                             # Anonymous call
```

### DIFF
--- a/dsl/scoped_executors.rb
+++ b/dsl/scoped_executors.rb
@@ -21,8 +21,10 @@ end
 
 execute do
   cmd(:before) { "echo '--> before'" }
-  call { :capitalize_a_random_word }
-  call { :capitalize_a_random_word }
-  call { :capitalize_a_random_word }
+  # First argument to `call` is the executor scope to run
+  # Second argument is the (optional) cog name
+  call(:capitalize_a_random_word, :first_call)
+  call(:capitalize_a_random_word, :other_named_call)
+  call(:capitalize_a_random_word) # anonymous call cog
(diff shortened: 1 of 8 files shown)


juniper-shopify:
### Merge activity

* **Oct 20, 7:11 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/463).

PR #467, opened 2025-10-20
Resolve plural step names to plural class names

### LINKED ISSUE #335: Plain ruby step de-pluralizes to get constant
```
Executing: generate_tests (Resource type: none)
Requiring step file: /Users/nathanfaber-good/world/trees/local-testgen-basic-functionality/src/areas/core/shopify/roast/ci-coverage/generate_tests.rb

❌ Step failed: 'generate_tests'
   Error: uninitialized constant GenerateTest
   This may be an issue with the step's implementation.
bundler: failed to load command: roast (/Users/nathanfaber-good/.dev/gem/5363350hi5f719harw2bnznq2gmjdilb-nix-shell/bin/roast)


🛑 Workflow stopped due to step failure.
   Check the error message above for details.

For debugging, you can run with --verbose for more details.
/Users/nathanfaber-good/.dev/gem/5363350hi5f719harw2bnznq2gmjdilb-nix-shell/bundler/gems/roast-48e5c8ef9f59/lib/roast/workflow/error_handler.rb:103:in 'Roast::Workflow::ErrorHandler#handle_generic_error': Failed to execute step 'generate_tests': uninitialized constant GenerateTest (Roast::Workflow::WorkflowExecutor::StepExecutionError)
```

alexanderadam:
fixes #335

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev
**PPS:** would you be so kind and add the `hacktoberfest-accepted` label to this PR in case you find that PR helpful? :pleading_face:

### DIFF
--- a/lib/roast/workflow/step_loader.rb
+++ b/lib/roast/workflow/step_loader.rb
@@ -169,7 +169,14 @@ module Roast
           raise StepExecutionError.new("Syntax error in step file: #{e.message}", step_name: step_name, original_error: e)
         end
 
-        step_class = step_name.classify.constantize
+        step_class = if Object.const_defined?(step_name.camelize, false)
+          step_name.camelize.constantize
+        elsif Object.const_defined?(step_name.classify, false)
+          step_name.classify.constantize
+        else
+          raise StepExecutionError.new("No class named #{step_name.camelize} found in step file #{file_path}", step_name: step_name)
+        end
+
         context = File.dirname(file_path)
         # For Ruby steps, we instantiate the specific class directly
         # Convert step_name to StepName value object
--- a/test/roast/workflow/step_loader_test.rb
+++ b/test/roast/workflow/step_loader_test.rb
@@ -295,6 +295,66 @@ module Roast
           assert_equal({ "CUSTOM_VAR" => "custom_value" }, step.env)
         end
       end
+
+      def test_loads_ruby_step_with_plural_name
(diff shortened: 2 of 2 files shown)


nfgrep:
Unfortunately there's a decent chance folks have already relied on the mis-matched pattern here, so we'd need a way to migrate out of it. This along with current codefreeze means we likely just won't tackle this.

nfgrep:
>Should there be a (deprecation) warning for the fallback?
Because in the (very unlikely) event that both exist, this could lead to surprising behaviour.

I think, if you can pull off the `classify -> camelize` fallback then I don't think we'd need a deprecation. If the old cases all still work then we should be good! :)

alexanderadam:
> > Should there be a (deprecation) warning for the fallback?
> > Because in the (very unlikely) event that both exist, this could lead to surprising behaviour.
> 
> I think, if you can pull off the `classify -> camelize` fallback […]

Currently the code is `camelize` -> `classify` which feel 'more natural' to me. Do you want to have it inverted?

juniper-shopify (changes requested):
Nice! I think the fallback logic will work nicely. Just an issue with the one test case.  I think changing it or adding one additional one to explicitly test the existing behaviour that @nfgrep mentioned:

> You can have a do_things step in workflow.yml, and in a do_things.rb have a DoThing, gave this a go last night and it worked.

Issue #267, opened 2025-06-18
Sessions file name too long


ericproulx:
File name should be the step name and not the prompt or content

<img width="1322" alt="Image" src="https://github.com/user-attachments/assets/57442691-ee18-455d-bd88-87229cf038cb" />

PR #468, opened 2025-10-20
Avoiding session file names that are too long

alexanderadam:
refs #267

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev

### DIFF
--- a/lib/roast/workflow/file_state_repository.rb
+++ b/lib/roast/workflow/file_state_repository.rb
@@ -6,6 +6,8 @@ module Roast
     # File-based implementation of StateRepository
     # Handles state persistence to the filesystem in a thread-safe manner
     class FileStateRepository < StateRepository
+      MAX_FILENAME_LENGTH = 255
+
       def initialize(session_manager = SessionManager.new)
         super()
         @state_mutex = Mutex.new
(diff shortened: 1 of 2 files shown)


lib/roast/workflow/file_state_repository.rb
```diff
@@ -90,9 +92,9 @@ def find_step_files(session_dir)
       end
 
       def find_step_before(step_files, target_step_name)
-        # First try to find if we have the exact previous step
         step_files.each_with_index do |file, index|
-          next unless file.end_with?("_#{target_step_name}.json")
+          state_data = load_state_file(file)
+          next if state_data[:step_name] != target_step_name && state_data[:step_name] != target_step_name.to_s
```
nfgrep:
```suggestion
          next if state_data[:step_name] != target_step_name.to_s
```
Do we need both checks here? In what case will `target_step_name` not be a String? If there is a case like that, why would we want to compare it to `state_data[:step_name]` 🤔

lib/roast/workflow/file_state_repository.rb
```diff
@@ -145,8 +147,61 @@ def copy_states_to_new_session(workflow, source_session_dir, state_files)
...
+        sanitized_parts.join("/")
+      end
+
+      # Safely truncate a string to a maximum byte length
+      # Ensures we don't cut in the middle of a multi-byte UTF-8 character
+      #
+      # @param str [String] The string to truncate
+      # @param max_bytes [Integer] Maximum byte length
+      # @return [String] The truncated string
+      def truncate_safely(str, max_bytes)
```
nfgrep: Do we need the manual byte manipulation here? I thought ruby handled utf8 strings on its own. Something like `str[0..max_len]` should do it I think.

nfgrep (changes requested):
Appreciate the move into file_state_repository.rb ❤️ Just a few straggler questions to answer, with those addressed we should be good to merge :)

PR #472, opened 2025-10-21
Refactor system cogs to use keyword arguments for custom params

juniper-shopify:
# Refactor system cogs to use keyword arguments for custom params

Pursuant to our discussion this morning @dersam and @nfgrep, this PR changes the way system cogs accept their custom parameters (e.g., for the `call` and `map` cogs, the name of the execution scope they should run).

### Design Principles
- The cog's `name` argument should always be first
- The name argument should always be optional
  - Therefore, system cogs cannot take required positional arguments
- Keyword arguments provide better clarity to readers of the workflow anyway

Example:
```
  call(:named_call, run: :some_subroutine)
  call(run: :some_subroutine) # anonymous call
```

### DIFF
--- a/dsl/map.rb
+++ b/dsl/map.rb
@@ -14,7 +14,7 @@ execute(:capitalize_a_word) do
     word ||= cmd(:word).out.strip
     my.command = "sh"
     my.args << "-c"
-    my.args << "echo '#{word}' | tr '[:lower:]' '[:upper:]'"
+    my.args << "echo \"#{word}\" | tr '[:lower:]' '[:upper:]'"
   end
 end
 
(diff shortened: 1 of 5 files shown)


juniper-shopify:
* **#472** <a href="https://app.graphite.dev/github/pr/Shopify/roast/472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

juniper-shopify:
I decided to call the keyword argument `run` -- since I think that makes it extra clear what it's going to do -- instead of the previous term `scope` I had been using. What do you think of that?

I think when we have a way to retrieve output from `subroutines`, like reducing the results of a map for instance, it might make more sense to refer to the thing that ran, from which we're retrieving output, as the `scope`. So we might want to revisit naming consistency when we have some prototype implementations of that to look at

PR #473, opened 2025-10-21
Add skip! and fail! control flow methods to cogs, and bang- and question- versions of output accessors

juniper-shopify:
This PR include two related concepts that are hard to separate in terms of implementation. I think we should discuss them together.

# Introduce `skip!`, and `fail!` methods for cog flow control

## skip!

Sometimes, you may have a reason to conditionally skip a cog, based on the output of a previous cog. The `skip!` method available in the cog's input proc will do this.

For instance, if you have a step that fetches some data that could be in several formats, you may need to transform it in some way, but only if it is in one of its possible formats. You could define a workflow like this in a simple and clean way:

(I'm using a pseudo-code cog named `cog` as a stand-in for any kind of cog; it'd doesn't matter to the example)

```
cog(:fetch_data) { "... fetches some data..." }

cog(:clean_data) do
  skip! if cog(:fetch_data).clean? 
  "... cleans the data..."
end
```

## fail!

You may also have a case where, because of some aspect of a prior cog's output (not necessarily a failure on its part, though), you don't have a valid state in which to continue the workflow. You could, of course, just raise any old exception, but having a `fail!` method that allows you to communicate to the workflow manager than the failure was sort of intentional allows us to better handle this kind of failure (we definitely don't need to show a stack trace, for example).

For instance, if you have a step that fetches data from an API, you might want to gracefully fail the workflow if the API returns a 404. This is a case you expect to encounter sometimes, though, so you want Roast to handle this kind of failure more gracefully. (TBD, though, what that 'gracefully' will actually look like, in terms of process exit code, what we print to the console, etc.) A workflow that uses this method might look something like this example.

```
cog(:fetch_api_data) { "... fetches some data from a web api..." }

cog(:process_data) do
  fail! unless cog(:fetch_api_data).response.status == 200
  " ... does something with the data ..."
end
```

# Add bang, non-bang, and question versions of cog output accessors

In a world where cogs can be skipped, and/or where cogs can fail but the workflow proceed, it becomes possible, in a subsequent cog's input proc, for the user to request the output of a previous cog that did not complete successfully, or to want to know if a previous cog did or did not run successfully

The current implementation of the cog output accessor methods (what I'm calling the cog name method you call inside a cog's input proc to get a prior cog's output) will raise an exception if the cog did not run and successfully produce output.

I think it's ugly and verbose to expect the user to rescue these exceptions in the input proc, especially if they need to do it a bunch. So, I've added bang and non-bang versions of the output accessor methods, along with a question-mark version that just returns a boolean if the cog's output is / is not present.

Pros:
* clean, clear syntax that is good Ruby style
* simple for user to decide "i want this to throw if the cog didn't run successfully" or "I want to handle the case where the cog didn't run successfully myself" or "I want to ask if this cog ran, but I don't need the output right now"
* `cog?(:name)` is clearer and more idiomatic that calling `cog(:name).nil?` / `cog(:name).present?` as a way to ask "did this cog run successfully?"

Cons:
* most of the time, the user probably wants the bang version of the method (though tbh I'm less sure now that this is true than when I first raised this PR)
* maybe the user still wants some of the output access exceptions to be raised (like, give me nil if the cog was skipped, but raise if the cog failed)

When I was working on this, I thought I was going to kind of hate having to use the bang methods most of the time, but when I actually updated the prototype workflows, I found out that I actually liked it. Curious what you all think.

(Also, for the second con, if the user wants some of the exceptions to still be raised, they can just use the bang method and rescue the ones that want to. I don't think we need a custom mechanism for that in the DSL).

UPDATE: The `CogDoesNotExistError` will *always* raise even with the non-bang or question-mark versions of the output accessor method. In such a case, the workflow is actually invalid, in a way that we would ideally be able to catch at compile time (but can't because the problematic statement is inside a cog input block that we can't evaluate until runtime).

### DIFF
--- a/dsl/skip.rb
+++ b/dsl/skip.rb
@@ -1,0 +1,36 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { print_all! }
+end
+
+execute do
+  cmd(:seconds) { "date +'%S'" }
+
+  cmd(:even) do
+    # `cmd!` output accessor bang method will raise an exception if the named cog did not complete successfully
+    seconds = cmd!(:seconds).out.strip.to_i
+    # Use the `skip!` method in a cog's input proc to conditionally skip this cog
+    # If `skip!` is called, the input proc will immediately terminate
+    skip! if seconds.odd?
+    "echo '#{seconds} is even'"
+  end
+
+  cmd(:odd) do
+    seconds = cmd!(:seconds).out.strip.to_i
+    skip! if seconds.even?
+    "echo '#{seconds} is odd'"
+  end
+
+  cmd do |my|
+    my.command = "echo"
+    # `cmd` output accessor non-bang method will return the cog's output or
+    # nil if the named cog did not run yet or did not complete successfully
+    my.args << "'even' cog ran" unless cmd(:even).nil?
+    # `cmd?` output accessor question method will return true/false if the cog did / did not complete successfully yet
+    my.args << "'odd' cog ran" if cmd?(:odd)
+  end
+end
--- a/dsl/call.rb
+++ b/dsl/call.rb
@@ -14,7 +14,7 @@ execute(:capitalize_a_random_word) do
   cmd(:capitalize) do |my, value_from_call|
     # Optional second block argument lets you use a value passed to the sub-executor by `call`
     # from withing any cog in the sub-executor
-    word = value_from_call || cmd(:word).out.strip
+    word = value_from_call || cmd!(:word).out.strip
     my.command = "/bin/sh"
     my.args << "-c"
     my.args << "/bin/echo \"#{word}\" | tr '[:lower:]' '[:upper:]'"
--- a/dsl/map.rb
+++ b/dsl/map.rb
@@ -11,7 +11,11 @@ end
 
 execute(:capitalize_a_word) do
   cmd(:capitalize) do |my, word|
-    word ||= cmd(:word).out.strip
+    # Use the `fail!` method in a cog's input proc to terminate the cog with a failure state
+    # e.g., if a condition prevents its successful execution.
+    # If `fail!` is called, the input proc will immediately terminate.
+    # The entire workflow may or may not terminate as a result, based on its configuration and the cog's configuration.
+    fail! unless word.present?
     my.command = "sh"
     my.args << "-c"
     my.args << "echo \"#{word}\" | tr '[:lower:]' '[:upper:]'"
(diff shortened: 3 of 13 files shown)


juniper-shopify:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/roast/473?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#473** <a href="https://app.graphite.dev/github/pr/Shopify/roast/473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#472** <a href="https://app.graphite.dev/github/pr/Shopify/roast/472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

dsl/skip.rb
```diff
@@ -0,0 +1,34 @@
...
+  cmd(:odd) do
+    seconds = cmd!(:seconds).out.strip.to_i
+    skip! if seconds.even?
+    "echo '#{seconds} is odd'"
+  end
+
+  cmd do |my|
+    my.command = "echo"
+    # `cmd` output accessor non-bang method will return nil if the named cog did not run yet or did not complete successfully
+    my.args << "'even' cog ran" unless cmd(:even).nil?
```
dersam: Instead of `nil?`, what about this:

```suggestion
    my.args << "'even' cog ran" unless cmd?(:even)
```

dersam (commented):
This general approach is in line with some of my existing thinking on control flow. I had already though we would need some version of `skip!` or `fail!` so no objections there.

Going to let this marinate in my head for a bit before responding more.

dsl/skip.rb
dersam: Yes, definitely. Any attempt to reference an invalid cog should fail loudly, since that should never be something you want to do.

juniper-shopify: done!

dersam (approved):

juniper-shopify:
### Merge activity

* **Oct 21, 6:34 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/473).
* **Oct 21, 6:35 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/473).

PR #480, opened 2025-10-22
Add `from` output accessor to get context from other execution scopes

juniper-shopify:
This is the first PR in a stack of new 'meta' output accessors to allow access to the outputs of cogs that were run in a subroutine via `call` or `map`, etc.

## Basic design paradigm

The output of the `call` and `map` cogs themselves do not themselves contain the output of the cogs that ran in their subordinate executor blocks. I considered this but determined that it gets really messy and/or gives the user limited flexibility when it comes to how they want to consume the output, and/or requires us to re-implement big chunks of the `CogInputManager`.

Instead, there should be a simple way to say "run this output accessor logic **in the context of the `CogInputContext` of the subordinate executor**, instead of in the `CogInputContext` of the current executor.

## Example

```
execute(:some_scope) do
  cog(:other) { ... does something ... }
end

execute do
  cog(:normal) { ... does something ... }
  call(:subroutine, run: :some_scope)
  other_cog {
    value_from_own_scope = cog(:normal).value
    value_from_subroutine_scope = from(:subroutine) { cog(:other).value }
  }
end
```

---

## Flow Diagram

This diagram illustrates how the `from` method enables cross-scope output access when writing Roast workflows. See
`dsl/from.rb` for a complete working example.

### The Problem: Scope Isolation

```
┌─────────────────────────────────────────────────────────────────┐
│ execute(:capitalize_a_word) do                                  │
│   cmd(:to_original) { |_, word| "echo \"#{word}\"" }            │
│   cmd(:to_upper) { ... }                                        │
│   cmd(:to_lower) { ... }                                        │
│ end                                                             │
└─────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────┐
│ execute do                                                      │
│   call(:hello, run: :capitalize_a_word) { "Hello" }             │
│                                                                 │
│   cmd do                                                        │
│     # ❌ ERROR: :to_upper doesn't exist in this scope           │
│     cmd(:to_upper)                                              │
│   end                                                           │
│ end                                                             │
└─────────────────────────────────────────────────────────────────┘
```

**Why?** Each `execute` block has its own isolated CogInputContext. The cogs defined in `:capitalize_a_word` are not
accessible from the main execute block.

### The Solution: Context Switching with `from`

```
┌─────────────────────────────────────────────────────────────────┐
│ execute do                                                      │
│   call(:hello, run: :capitalize_a_word) { "Hello" }             │
│                                                                 │
│   cmd do                                                        │
│     # ✅ Works: `from` switches to :hello's context             │
│     upper = from(call!(:hello)) { cmd!(:to_upper).out.strip }   │
│   end                                                           │
│ end                                                             │
└─────────────────────────────────────────────────────────────────┘
```

### Execution Stack Detail

From `dsl/from.rb` lines 41-44:

```ruby
cmd do
  original = from(call!(:hello)) { cmd!(:to_original).out.strip }
  upper = from(call!(:hello)) { cmd!(:to_upper).out.strip }
  lower = from(call!(:hello)) { cmd!(:to_lower).out.strip }
  "echo \"#{original} --> #{upper} --> #{lower}\""
end
```

**Visual Flow:**

```
Main CogInputContext
│
│  You are here (inside cmd block)
│  Available: call(:hello), call(:world)
│  NOT available: cmd(:to_upper), cmd(:to_lower)
│
├─ Call: from(call!(:hello)) { cmd!(:to_original).out.strip }
│   │
│   ├─ Step 1: call!(:hello) retrieves the Call output
│   │           └─> This output "remembers" its CogInputContext
│   │
│   ├─ Step 2: from(...) switches context
│   │           │
│   │           ▼
│   │  ┌────────────────────────────────────────────────┐
│   │  │ :capitalize_a_word CogInputContext             │
│   │  │                                                │
│   │  │ Available: cmd(:to_original)                   │
│   │  │            cmd(:to_upper)                      │
│   │  │            cmd(:to_lower)                      │
│   │  │                                                │
│   │  │ Execute: cmd!(:to_original).out.strip          │
│   │  │           └─> Returns: "Hello"                 │
│   │  └────────────────────────────────────────────────┘
│   │           │
│   │           ▼
│   └─ Step 3: Return to main context with result
│              └─> original = "Hello"
│
├─ Call: from(call!(:hello)) { cmd!(:to_upper).out.strip }
│   └─> Switches context again → returns "HELLO"
│       upper = "HELLO"
│
├─ Call: from(call!(:hello)) { cmd!(:to_lower).out.strip }
│   └─> Switches context again → returns "hello"
│       lower = "hello"
│
└─ Back in main context, build command string
    "echo \"Hello --> HELLO --> hello\""
```

### DIFF
--- a/dsl/from.rb
+++ b/dsl/from.rb
@@ -1,0 +1,55 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { display! }
+  cmd(/to_/) { no_display! }
+end
+
+execute(:capitalize_a_word) do
+  cmd(:to_original) { |_, word| "echo \"#{word}\"" }
+  cmd(:to_upper) do |my, word|
+    my.command = "sh"
+    my.args << "-c"
+    my.args << "echo \"#{word}\" | tr '[:lower:]' '[:upper:]'"
+  end
+  cmd(:to_lower) do |my, word|
+    my.command = "sh"
+    my.args << "-c"
+    my.args << "echo \"#{word}\" | tr '[:upper:]' '[:lower:]'"
+  end
+end
+
+execute do
+  # Call a subroutine with `call`
+  call(:hello, run: :capitalize_a_word) { "Hello" }
+  call(:world, run: :capitalize_a_word) { "World" }
+
+  cmd do
+    # Normally, you can only reference the output of cogs that run in the same executor scope.
+    begin
+      # There is no :to_upper cog that this could sensibly be referring to
+      cmd(:to_upper)
+    rescue Roast::DSL::CogInputManager::CogDoesNotExistError
+      puts "Could not access :to_upper directly"
+    end
+
+    # Using `from`, you can access cogs from the executor scope that was run by a specific named `call`.
+    # The block you pass to `from` runs in the input context of the specified scope, rather than the current scope.
+    original = from(call!(:hello)) { cmd!(:to_original).out.strip }
+    upper = from(call!(:hello)) { cmd!(:to_upper).out.strip }
+    lower = from(call!(:hello)) { cmd!(:to_lower).out.strip }
+    "echo \"#{original} --> #{upper} --> #{lower}\""
+  end
+
+  cmd do
+    # You can also grab the `call`'s output once and pass it to multiple `from` invocations.
+    my_scope = call!(:world)
+    original = from(my_scope) { cmd!(:to_original).out.strip }
+    upper = from(my_scope) { cmd!(:to_upper).out.strip }
+    lower = from(my_scope) { cmd!(:to_lower).out.strip }
+    "echo \"#{original} --> #{upper} --> #{lower}\""
+  end
+end
--- a/lib/roast/dsl/cog_input_context.rb
+++ b/lib/roast/dsl/cog_input_context.rb
@@ -5,6 +5,11 @@ module Roast
   module DSL
     # Context in which the individual cog input blocks within the `execute` block of a workflow definition are evaluated
     class CogInputContext
+      include SystemCogs::Call::InputContext
+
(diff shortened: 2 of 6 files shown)


juniper-shopify:
Not one specific thing gets specifically messy, but definitely a lot of extra boilerplate you'd need to write, for one thing. And we lose a lot of the benefits of the `CogInputContext` and `CogInputManager` structures we've set up. Cogs produce various amounts of output, only some of which is needed in subsequent steps. If we want to access everything from the subroutine via the output of the `call` cog, then you either need to:
* A) write a bunch of code (somewhere, there isn't an obvious place for it now) to extract all of the bits of cog output you want and combine them into a subroutine-level output object, which you'd also need to define the shape of or deal with an untyped hash. If you decide somewhere else you want to use some additional piece of output from the subroutine, you need to go pipe it through from the top. Or,
* B) we jam all the output of every cog that runs in a subroutine onto the call cog's output. It's basically impossible to type this object correctly, since its members will be various different kinds of cogs' outputs, identified by name, which are only set in the workflow itself. Doing this gets really messy when there are calls nested within calls (which I'm sure will happen for someone at some point).

Doing either of these (especially B) also basically defeats all the value we currently get from the `CogInputContext` and `CogInputManager`, which present for you a nice interface for retrieving cog output. You'd get the nice interface for output from cogs in the current scope, but to get output from cogs in a subordinate scope, you'd just have a giant object with everything mushed together.

Either of these options would also make the `collect` and `reduce` operations for `map` calls grosser. You could well make conditional choices to access certain bits of cog output from one instance of a scope than you do from another, based on some third cog's output, or whether a cog failed or was skipped in each run.

It also means we need to reinvent the semantics of 'did this cog run?', or 'throw if the cog whose output I want did not actually run'

PR #483, opened 2025-10-22
Add `reduce` output accessor method to reduce results of running a `map` cog

juniper-shopify:
This PR adds `reduce` output accessor method that mirrors the functionality of `collect` and `from` in downstack PRs.

The `reduce` output accessor has generally the same semantics as the normal Ruby `reduce` method on an iterable. You give it an optional initial_value, and a block that takes an accumulator as argument. The return value of the block becomes the value of the accumulator passed to the block the next time it is called, and the return value of the block the final time it is invoked becomes the return value of `reduce` itself.

There is one key semantic difference. Ruby's normal reduce, called on an iterable of objects, takes a block that accepts two arguments, the accumulator and each object from the iterable in turn. The `reduce` output accessor here only passes one argument, the accumulator, to its block. Instead, the block is executed __in the context of each execution scope from the map being reduced__. Thus, no second 'item' argument is necessary.

### DIFF
--- a/dsl/map_reduce.rb
+++ b/dsl/map_reduce.rb
@@ -1,0 +1,37 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { display! }
+  cmd(/to_/) { no_display! }
+end
+
+execute(:capitalize_a_word) do
+  cmd(:to_original) { |_, word| "echo \"#{word}\"" }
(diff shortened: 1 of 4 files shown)


juniper-shopify:
### Merge activity

* **Oct 22, 2:30 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/483).

PR #486, opened 2025-10-24
Add shorthand syntax for `from`

juniper-shopify:
Adds a shorthand version of `from` when you just want the output of the last cog in the sub executor scope:

__Existing syntax (retained):__
```
from(call(:foo)) { cog(:name) } # (with a block)
```
gets you the output of the block you pass, evaluated in the subroutine’s `CogInputContext`

__New syntax (added):__
```
from(call(:foo)) # (no block given)
```
gets you the output of the last cog in the subroutine (just like you’d expect the return value of a ruby method to be)

### DIFF
--- a/lib/roast/dsl/system_cogs/call.rb
+++ b/lib/roast/dsl/system_cogs/call.rb
@@ -69,12 +69,18 @@ module Roast
 
         # @requires_ancestor: CogInputContext
         module InputContext
-          #: [T] (Roast::DSL::SystemCogs::Call::Output) {() -> T} -> T
+          # @rbs [T] (Roast::DSL::SystemCogs::Call::Output) {() -> T} -> T
+          #    | (Roast::DSL::SystemCogs::Call::Output) -> untyped
           def from(call_cog_output, &block)
             em = call_cog_output.instance_variable_get(:@execution_manager)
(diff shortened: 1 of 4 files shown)


lib/roast/dsl/system_cogs/call.rb
```diff
@@ -69,12 +69,18 @@ def create_call_system_cog(params, input_proc)
 
         # @requires_ancestor: CogInputContext
         module InputContext
-          #: [T] (Roast::DSL::SystemCogs::Call::Output) {() -> T} -> T
+          # @rbs [T] (Roast::DSL::SystemCogs::Call::Output) {() -> T} -> T
```
dersam: What's this `@rbs`?

dersam (changes requested):

lib/roast/dsl/system_cogs/call.rb
dersam: TIL

juniper-shopify: hmm, it does seem like sorbet doesn't actually parse it, but RubyMine does, so I have to add the .rbi shim as well so sorbet will know about it in when you're writing a workflow. But, there's no other way I've been able to find besides this to have any valid version of a type annotation here on this method at all.

PR #487, opened 2025-10-24
Add shorthand syntax for `collect`

juniper-shopify:
Adds a shorthand version of `collect` when you just want the outputs of the last cog in the sub executor scopes run by `map`:

__Existing syntax (retained):__
```
collect(map(:foo)) { cog(:name) } # (with a block)
```
gets you an Array of the outputs of the block you pass, evaluated in the each of the map subroutine’s `CogInputContext`s

__New syntax (added):__
```
collect(map(:foo)) # (no block given)
```
gets you an Array of the outputs of the last cog in each of the map's subroutines (similar to the new shorthand version of `from`)


juniper-shopify:
### Merge activity

* **Oct 24, 5:29 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/487).

PR #495, opened 2025-10-27
Add async gem and clean up tapioca RBIs

juniper-shopify:
I'm adding the `async` gem in preparation for adding async execution support to `map`. 

Doing this, I had to update / clean up some tapioca-generated file, so I'm splitting that stuff into it's own PR to make the actual async map PR easier to review.


juniper-shopify:
### Merge activity

* **Oct 29, 8:39 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/495).

juniper-shopify:
`async` uses Fibers instead of Threads. Ractors use threads, but (TIL!) with some cool optimizations to achieve actual multi-process-like parallelization across CPUs, which you can't do with native Threads.

My intuition right now is that we'll mostly be IO-bound, running LLM tools with the `chat` and `agent` cogs -- and even shell commands we run with the `cmd` cog are separate external processes -- so Fibers will be lower overhead at least. (And especially where our architecture theoretically allows parallelization inside parallelization, I think a coöperative multitasking approach is simpler)

For what we're doing right now, I don't think there's a huge difference what framework we choose, though. I'd be down for weaving in Ractors in the very specific case that we build a cog that explicitly wants to do parallel in-process work on a large dataset, in Ruby.

PR #496, opened 2025-10-27
Add optional parallelism to `map` cog

juniper-shopify:
This PR makes it possible to use `map` to optionally process its items in parallel.

Syntax is:

```
config do
  map(:foo) do
    parallel! # allows unlimited parallelism; all items begin processing at the same time
  end

  map(:bar) do
    parallel(3) # allows up to 3 items to be processed in parallel at the same time.
  end
end
```

The items will begin executing in the order they were passed to `map`, but may complete in any order. When the output of the map is accessed, results will always be returned in the same order in which their respective inputs were given to `map`.

### DIFF
--- a/dsl/parallel_map.rb
+++ b/dsl/parallel_map.rb
@@ -1,0 +1,37 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { display! }
+  map(:words) do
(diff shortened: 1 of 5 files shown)


dersam (approved):

juniper-shopify:
### Merge activity

* **Oct 29, 8:44 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/496).

PR #497, opened 2025-10-27
Allow generic cogs to be run asynchronously

juniper-shopify:
This PR adds the ability trigger async processing for any cog.

The motivation for this feature is somewhat distinct from async `map`s. In the map case, it's basically "process these items in parallel; but don't move on to the next cog in the overall stack until all items are processed". With individual async cogs, it's basically "fork this one cog and let it run in the background while execution continues down the cog stack"

One would generally do something like this:

```
config do
  cmd(:slow) do
    async!
  end
end

execute do
  cmd(:slow) { "kick off something to run in the background" }
  cmd(:foo) { "this does not depend on :slow, so it can start right away" }
  cmd(:bar) { "ditto for this one; it runs after :foo completes, like normal" }
  cmd(:use_output_of_slow) do
    result = cmd!(:slow).out
    "this one depends on the output of :slow, so it will block until :slow is done"
  end
  cmd(:baz) { "this does not need :slow, but it runs synchronously after :use_output_of_slow, so it doesn't start until that cog is done; like normal" }
end
```

### DIFF
--- a/dsl/async_cogs.rb
+++ b/dsl/async_cogs.rb
@@ -1,0 +1,49 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { print_all! }
+
+  # Any type of cog can be configured to run asynchronously in the background.
+  # The cog that follows it in the workflow will be able to begin running immediately.
+  cmd(/slow/) { async! }
+end
+
+execute do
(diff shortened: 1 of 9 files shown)


juniper-shopify:
### Merge activity

* **Oct 29, 8:50 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/497).

PR #498, opened 2025-10-27
Add a `ruby` cog that just outputs the return value of the ruby code you write in its input block

juniper-shopify:
This PR adds a `ruby` cog that just outputs the return value of the ruby code you write in its input block

This is a more powerful and flexible replacement for the Ruby step functionality in legacy Roast.

---

__You can pretty easily do any of the following__

In-line Ruby code:
```
execute do
  ruby(:foo) do
    # write any Ruby code you want
    # access the output of previous cogs; whatever
    # return anything you want
  end

  cmd do
    # return value of `ruby(:foo)` is available as the `value` attribute on its output object
    results = ruby(:foo).value
  end
end
```

external Ruby class step
```
require "path/to/something.rb"

execute do
  ruby(:foo) do
    something = Something.new("whatever")
    something.do_a_thing
  end
end
```

in-line Ruby, defined elsewhere in the workflow file:

```
# note, you need to put your method in a class, because the `basic` cog's block runs in a different context
class MyClass
  class << self
    def my_method(a, b)
      # write some code here
  end
end

execute do
  ruby(:foo) { MyClass.my_method(1, 2) }
  ruby(:bar) { MyClass.my_method(3, 4) }
end
```

### DIFF
--- a/dsl/ruby_cog.rb
+++ b/dsl/ruby_cog.rb
@@ -1,0 +1,37 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+class MyClass
+  class << self
+    def add_stuff(a, b)
+      a + b
+    end
+  end
+end
+
+config do
+  cmd { print_all! }
+end
+
+execute do
+  cmd(:roast) { "echo Roast" }
+
+  # The `ruby` cog just passes the return value from its input block directly to its output.
+  # Letting you write whatever ruby code you want to generate that output
+  ruby(:whatever) do
+    # Do whatever you want in this block
+    value = cmd!(:roast).out
+    puts "Hello, #{value.upcase}"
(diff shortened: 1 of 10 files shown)


dersam (changes requested):
Can we call the cog `ruby` instead? `basic` doesn't really tell me what it does.

juniper-shopify:
Absolument!

That was my first thought, actually, but one concern I had with that is it shifts the user's perspective to thinking of the cog input block as "where I write the code that does the work of the cog" (that is, for the `ruby` cog, the ruby code you write for the cog's execution is in the input block). I would prefer people think of the input block as "where I write the code that sets up the input for the cog's execution".  By describing it as "a basic no-op cog that passes its input to its output" I was trying to mitigate that.

If you're not worried about people getting confused by this, then I'm 100% on board with renaming it. I do like `ruby` better.

PR #499, opened 2025-10-28
Add validator block to field definition, and explicit use_default bang method

juniper-shopify:
This PR adds an optional validator block to the `field` definition functionality on cog input.
The validator, if present, can raise an error and optionally coerce the new user-provided value to
a canonical form.

---

This PR also provides a clearer way for users to explicitly indicate in their config blocks that they wish to explicitly use the default value of a config field, overriding a previously configured non-default value.

The field can be invoked in three ways:

```
config do
  agent do

  # sets the provider to :something
  provider :something

  # Sets the provider to be the configured default value for the field
  # NOTE: this sets the default explicitly, meaning that it will overwrite the value selected on a more generic config level
  use_default_provider!

  # Returns the current value of the provider, without changing it
  # (previously, this would have reset the provider to the default value)
  provider

  end
end
```

The utility of explicitly setting the default can be demonstrated in an example of some config-merging behaviour:

```
config do

  agent do
    provider :my_provider
    model "my-model"
  end

  agent(:special) do
    provider :special_provider
    model "special-model"
  end

  agent(:normal) do
    use_default_provider!
  end

  agent(:other) do
    model "other-model"
  end
end

execute do

  # uses :my_provider and "my-model" (set in the global `agent` config block)
  agent { ... }

  # uses :my_provider and "my-model" (set in the global `agent` config block)
  agent(:random_name) { ... }

  # uses :special_provider and "special-model" (set in the :special `agent` config block)
  agent(:special) { ... }

  # uses the Roast default provider (overriding the global `agent` config block)
  # and "my-model" (set in the global `agent` config block)
  agent(:normal) { ... }

  # uses :my_provider (set in the global `agent` config block)
  # and "other-model" (set in the :other `agent` config block)
  agent(:other) { ... }
end
```

PR #500, opened 2025-10-28
Better nil handling for cog input objects

juniper-shopify:
# Better nil handling for cog input objects

This PR introduces a (slightly) cleaner way for cogs to get validated values from their input objects (when, for instance, a part of the validation is to reject `nil`s.

__The problem:__

An input defined like this can theoretically guarantee that `name` is never `nil` when the cog is executing (`validate!` gets called by Roast when the cog's input block is executed, before the cog's `run!` method is called).
```
class Input
  #: String?
  attr_accessor :name

  def validate!
    raise Cog::Input::InvalidInputError if name.nil?
  end
end
```

But when the cog is actually running and wants to use its input values, the type checker doesn't know they can't be `nil` (or whatever) at that point. So, we either need to re-implement (some of) the validation, or do some theoretically-safe-but-gross casts. No good!

```
def execute
  actual_name = config.name  #: as String
end
```

--- 

__The improvement:__

Pull steps of the `validate!` method out into their own methods that return a valid value if their part of the input is valid, or raise an error if it isn't.

```
class Config
  #: String?
  attr_accessor :name

  def validate!
    valid_name!
  end

  def valid_name!
    raise Cog::Input::InvalidInputError if @name.nil?
    @name
  end
end
```

so now:

```
def execute
  config.valid_name!  # this is already a String
end
```

---

Just a small tweak to a pattern for now that makes a small improvement. As we build out more cogs and see more use cases, maybe a better pattern will emerge. I also think we could probably derive a clever generator for these validators, similar to the `field` helper class method on `Cog::Config`.

### DIFF
--- a/lib/roast/dsl/cogs/agent.rb
+++ b/lib/roast/dsl/cogs/agent.rb
@@ -33,7 +33,7 @@ module Roast
 
           #: () -> void
           def validate!
-            raise Cog::Input::InvalidInputError, "'prompt' is required" unless prompt.present?
+            valid_prompt!
           end
 
           #: (untyped) -> void
@@ -41,6 +41,13 @@ module Roast
             if input_return_value.is_a?(String)
               self.prompt ||= input_return_value
             end
+          end
+
+          #: () -> String
+          def valid_prompt!
+            raise Cog::Input::InvalidInputError, "'prompt' is required" unless @prompt.present?
+
+            @prompt
           end
         end
 
@@ -60,27 +67,20 @@ module Roast
 
         #: (Input) -> Output
         def execute(input)
-          prompt = input.prompt
-          raise MissingPromptError, "Prompt is required for agent cog" unless prompt.present?
-
-          provider_name = config.provider
-          raise MissingProviderError, "Provider config is required for agent cog" unless provider_name.present?
-
-          provider_instance = create_provider(provider_name)
-          response = provider_instance.invoke(prompt)
+          response = provider.invoke(input.valid_prompt!)
           puts "[AGENT RESPONSE] #{response}"
           Output.new(response)
         end
 
         private
 
-        #: (Symbol) -> Roast::DSL::Cogs::Agent::Providers::Base
-        def create_provider(provider_name)
-          case provider_name
+        #: () -> Providers::Base
+        def provider
+          @provider ||= case config.provider
           when :claude
             Providers::Claude.new
           else
-            raise UnknownProviderError, "Unknown provider: #{provider_name}"
+            raise UnknownProviderError, "Unknown provider: #{config.provider}"
           end
         end
       end


juniper-shopify:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/roast/500?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#500** <a href="https://app.graphite.dev/github/pr/Shopify/roast/500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#499** <a href="https://app.graphite.dev/github/pr/Shopify/roast/499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#498** <a href="https://app.graphite.dev/github/pr/Shopify/roast/498?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#485](https://github.com/Shopify/roast/pull/485) <a href="https://app.graphite.dev/github/pr/Shopify/roast/485?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#497** <a href="https://app.graphite.dev/github/pr/Shopify/roast/497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#496** <a href="https://app.graphite.dev/github/pr/Shopify/roast/496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#495** <a href="https://app.graphite.dev/github/pr/Shopify/roast/495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#494** <a href="https://app.graphite.dev/github/pr/Shopify/roast/494?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#493** <a href="https://app.graphite.dev/github/pr/Shopify/roast/493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#492** <a href="https://app.graphite.dev/github/pr/Shopify/roast/492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#491** <a href="https://app.graphite.dev/github/pr/Shopify/roast/491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#490** <a href="https://app.graphite.dev/github/pr/Shopify/roast/490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

juniper-shopify:
### Merge activity

* **Oct 29, 9:17 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/500).

PR #501, opened 2025-10-29
Reorder Config, Params, Input, Output classes in Cog definitions

juniper-shopify:
We've ended up with a convention where the inner classes defined in a cog class
are given in the order Params (for system cogs), Input, Output, then Config.

This PR moves the Config class definitions to the top of each cog class.
This keeps the layout consistent across all cogs, but also puts these inner
classes in the position corresponding to which they are evaluated when the workflow
is executed.

### DIFF
--- a/lib/roast/dsl/system_cogs/call.rb
+++ b/lib/roast/dsl/system_cogs/call.rb
@@ -5,6 +5,8 @@ module Roast
   module DSL
     module SystemCogs
       class Call < SystemCog
+        class Config < Cog::Config; end
+
         class Params < SystemCog::Params
           #: Symbol
           attr_accessor :run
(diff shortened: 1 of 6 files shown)


lib/roast/dsl/system_cogs/call.rb
```diff
@@ -5,6 +5,8 @@ module Roast
   module DSL
     module SystemCogs
       class Call < SystemCog
+        class Config < Cog::Config; end
+
```
juniper-shopify: even though it is not necessary for a cog to define its own input class, I've come to the conclusion that it is easier and cleaner for the core Roast cogs to do so. It makes the experience of looking at the various object types more consistent, and it requires less changing of method signatures when we do decide that a previously-configured cog needs to have some config

PR #503, opened 2025-10-30
Add a marker for slow tests

juniper-shopify:
This PR adds a `slow_test!` helper method for tests that will skip the test unless is is being run via the `:minitest_dsl_slow` rake task (or with the 'ROAST_RUN_SLOW_TESTS' environment variable set).

CI is configured to run the fast tests first, and then the slow tests, (so we'll see errors that brea
the majority of test cases quickly) and the default behaviour of `bundle exec rake` will only
run the fast tests, making it easier for quick local dev iteration.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,9 +32,9 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-      - run: bundle exec rake rubocop_ci
       - run: bundle exec rake sorbet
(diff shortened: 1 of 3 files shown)


test/roast/dsl/command_runner_test.rb
```diff
@@ -104,6 +105,7 @@ class CommandRunnerTest < ActiveSupport::TestCase
       end
 
       test "timeout kills the process" do
+        slow_test!
```
juniper-shopify: ditto from above comment

dersam (approved):

PR #511, opened 2025-10-31
Add a `global` config scope to set basic things that apply to all cogs of all types

juniper-shopify:
Some config parameters, like 'exit on error', 'async' and 'working directory', are generally applicable to all/most cogs.
It makes sense to have a consistent definition for their setter/getter methods.

Many of these definitions exist on the Cog::Config base class, but user still need to specify them multiple times for each cog type.

Now you can set some options globally using a single block

```
config do
  global do
    # every cog of every time will run with this attribute set, unless overridden by a more specific config specification
    exit_on_error!
  end

  cmd do
    # overrides for all `cmd` cogs
    no_exit_on_error!
  end

  chat(:foo) do
    # overrides for a specifically named `chat` cog
    no_exit_on_error!
  end
end
```

### DIFF
--- a/dsl/outputs.rb
+++ b/dsl/outputs.rb
@@ -4,6 +4,7 @@
 #: self as Roast::DSL::Workflow
 
 config do
+  global { exit_on_error! }
   cmd { display! }
   cmd(/to_/) { no_display! }
 end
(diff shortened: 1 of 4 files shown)


juniper-shopify:
### Merge activity

* **Nov 3, 4:11 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/511).

PR #516, opened 2025-11-03
Run cogs in custom working directory

juniper-shopify:
Cogs that run external commands, like `cmd` and `agent`, or that access files on disk (like `chat`, when we give it the ability to do thing like handle image input/output) may benefit from being configured with an alternate working directory.

---

Note: This PR does not make Roast automatically change into the configured working directory when running a cog for which it is specified, because doing so causes issues for parallelization. (If two cogs are running in parallel and each attempts to change the working directory of
the main ruby process, then it gets changed for all parallel cogs, leading to unexpected behaviour. There's no way to effectively coördinate this.)

Instead cogs that need to interact with the filesystem, or care about their working directory for any reason, should use the `working_directory` config value internally to intentionally do whatever they want to do within that folder *without* changing the working directory of the ruby process.

(We'll maybe want to put some guardrails in place, at least a warning, to prevent the user from inadvertently changing the working directory in, say, a cog's input block. Though that's only an issue if they do use async cogs and don't realize/anticipate the side effects. Probably a niche issue.)

### DIFF
--- a/dsl/working_directory.rb
+++ b/dsl/working_directory.rb
@@ -1,0 +1,16 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::DSL::Workflow
+
+config do
+  cmd { print_all! }
+  cmd(:alt) { working_directory "/tmp" }
+  cmd(:orig) { use_current_working_directory! }
+end
+
+execute do
+  cmd(:cwd) { "echo Current working directory: `pwd`" }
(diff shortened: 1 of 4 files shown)


dersam (approved):

juniper-shopify:
### Merge activity

* **Nov 3, 4:51 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Shopify/roast/516).

PR #523, opened 2025-11-03
Add shell command support to CommandRunner (fixes #489)

### LINKED ISSUE #489: Add shell command support to CommandRunner
## Context

PR #485 adds CommandRunner with direct command execution (no shell features). This works for simple commands but doesn't support pipes, redirects, variable expansion, or other shell features that users expect.

## Related PR Comments

This issue addresses feedback from PR #485:

**Shell Support** - [@juniper-shopify (comment)](https://github.com/Shopify/roast/pull/485#discussion_r2460409545):
> "We don't need to implement this now in this PR -- it can get stacked on top"

Proposed API design with `execute()` for shell execution and `simple_execute()` for direct execution.

**User Expectations** - [@dersam (comment)](https://github.com/Shopify/roast/pull/485#discussion_r2460528950):
> "Agreed, you shouldn't have to think about the semantics of 'can I use pipes'. We already let you chain commands like a terminal by default in legacy Roast, and that's a good feature that we should continue to support."

**Process Cleanup** - [@juniper-shopify (comment)](https://github.com/Shopify/roast/pull/485#discussion_r2460369908):
> "I agree that we need this functionality. But I'm a little confused why the OS won't just give us this for free"

And [@juniper-shopify (comment)](https://github.com/Shopify/roast/pull/485#discussion_r2460451194):
> "`cleanup_all_children` is actually never called by anything..."

Clarified that:
- OS handles Ctrl+C cleanup for direct execution (single child)
- Timeout cleanup IS needed (implemented in #485)
- Full process tracking (hash, mutex, cleanup_all_children) only needed for shell execution where grandchildren can be orphaned
- Signal handlers needed to wire cleanup into DSL architecture

## Goal

Support shell features (pipes, redirects, etc.) as the primary API, making CommandRunner behave like a terminal.

## Proposed API

Per @juniper-shopify's suggestion in PR #485:

```ruby
# Primary API - shell execution (rename current execute)
CommandRunner.execute("ls -la | grep foo | awk '{print $1}'")
# Wraps with: sh -c "..."

# Power user API - direct execution (current implementation)
CommandRunner.simple_execute("ls", "-la")
# Direct: Open3.popen3("ls", "-la")
```

### 1. Shell Execution Method
- New `execute(command_string)` method that wraps commands with `sh -c "..."`
- Rename current `execute(*args)` to `simple_execute(*args)`
- Support all shell features: pipes, redirects, variable expansion, globbing

### 2. Process Group Management
Direct execution can kill the immediate child, but shell execution creates process trees:
```
roast → sh → sleep | grep (grandchildren)
```

When roast terminates (Ctrl+C or timeout), grandchildren become orphaned without process groups.

**Required:**
- Use `pgroup: true` in `Open3.popen3`
- Track process group ID (PGID)
- Kill entire process group: `Process.kill("-TERM", pgid)`

### 3. Enhanced Process Tracking
Current implementation has minimal timeout cleanup (PID tracking, kill_process helper). Shell support requires:
- Hash to track all child processes
- Mutex for thread-safe access
- `cleanup_all_children` method for graceful shutdown
- Track both PID and PGID for each process

### 4. Signal Handler Integration
Wire up signal handlers in DSL architecture:
- Trap INT (Ctrl+C) and TERM signals
- Call `cleanup_all_children` on signals
- Ensure proper cleanup on abnormal termination

### 5. Update Cmd Cog
- Determine when to use `execute` vs `simple_execute`
- Default to shell execution for user-facing commands
- Consider adding `shell:` config option

## Testing

- Shell pipelines: `echo foo | grep foo`
- Redirects: `echo test > /tmp/out.txt`
- Variable expansion: `echo $HOME`
- Process cleanup: verify grandchildren are killed on timeout/interrupt
- Signal handling: verify cleanup on Ctrl+C

## Dependencies

None - can be stacked on top of PR #485

## Related

- PR #485: Initial CommandRunner implementation
- Deferred from #485: Full process tracking, process groups, signal handlers

mathiusj:
### TL;DR

Restores shell command support (pipes, redirects, etc.) that was lost in PR #485, implementing Juniper's dual API design from issue #489.

### What changed?

**Core Implementation:**
- `execute(String)` - Primary API for shell commands with pipes, redirects, variable expansion
  - Wraps commands with `sh -c` under the hood
- `simple_execute(*args)` - Power user API for direct execution without shell
  - Safer for untrusted input, no shell injection risk

**Integration:**
- `cmd` cog automatically routes based on syntax:
  - `cmd { "git status | wc -l" }` → shell execution via `execute()`
  - `cmd { ["git", "status"] }` → direct execution via `simple_execute()`
- Both methods support `working_directory` parameter

**Examples & Documentation:**
- 4 runnable DSL examples demonstrating both approaches
- Complete documentation with security guidelines
- When-to-use guide for choosing between shell and direct execution

### How to test?

**Shell features (restored):**
```ruby
cmd(:count_files) { "git status --porcelain | wc -l" }
cmd(:with_vars) { "VAR='hello' && echo $VAR" }
cmd(:redirect) { "echo 'test' > /tmp/out.txt" }
```

**Direct execution (safe):**
```ruby
cmd(:safe) { ["echo", user_input] }  # No shell injection
cmd(:git) { ["git", "status", "--porcelain"] }
```

**Run examples:**
```bash
bin/roast execute dsl/shell_commands.rb --executor dsl
bin/roast execute dsl/direct_execution.rb --executor dsl
```

### Why make this change?

**Problem:** PR #485 introduced CommandRunner but only supported array syntax, breaking shell features that users expect (pipes, redirects, command substitution).

**Solution:** Implements the dual API design proposed by @juniper-shopify in #489:
- Restores shell support as the primary API
- Provides explicit direct execution for security-conscious use cases
- Maintains backward compatibility via smart routing in cmd cog

**Benefits:**
- ✅ Shell features work again (pipes, redirects, variables)
- ✅ Security: Array syntax prevents shell injection
- ✅ Clear separation: Method names indicate shell vs direct
- ✅ Documented: Examples show when to use each approach

Closes #489

### DIFF
--- a/dsl/README_CMD_EXAMPLES.md
+++ b/dsl/README_CMD_EXAMPLES.md
@@ -1,0 +1,114 @@
+# CMD Cog Examples
+
+These examples demonstrate the `cmd` cog's dual execution modes: shell commands and direct execution.
+
+## Running Examples
+
+```bash
+bin/roast execute dsl/shell_commands.rb --executor dsl
+bin/roast execute dsl/direct_execution.rb --executor dsl
+bin/roast execute dsl/shell_vs_direct.rb --executor dsl
+bin/roast execute dsl/working_directory_with_both.rb --executor dsl
+```
+
+## Examples Overview
+
+### 1. `shell_commands.rb` - Shell Features
+
+Demonstrates shell command execution with:
+
+- Pipes: `git status | wc -l`
+- Command substitution: `$(git branch)`
+- Variable expansion: `$VAR`
+- Redirects: `> /tmp/file.txt`
+- For loops and complex shell syntax
+
+**When to use:** You need shell features like pipes, redirects, or command substitution.
+
+### 2. `direct_execution.rb` - Safe Execution
+
+Demonstrates direct execution (array syntax) with:
+
+- No shell interpretation
+- Safe handling of special characters
+- Protection from shell injection
+- Literal argument passing
+
+**When to use:** You have untrusted input or want to avoid shell interpretation.
+
+### 3. `shell_vs_direct.rb` - Comparison
+
+Side-by-side comparison showing:
+
+- When to use shell (string syntax)
+- When to use direct (array syntax)
+- Security considerations
+- Performance implications
+
+**When to use:** Learning which approach fits your use case.
+
+### 4. `working_directory_with_both.rb` - Working Directory
+
+Demonstrates `working_directory` config with:
+
+- Shell commands in different directories
+- Direct execution in different directories
+- Per-command directory scoping
+
+**When to use:** You need to run commands in specific directories.
+
+## Syntax Quick Reference
+
+### Shell Command (String)
+
+```ruby
+cmd(:my_cmd) { "git status | wc -l" }
+```
+
+- ✅ Pipes, redirects, variables work
+- ⚠️ Shell injection risk with untrusted input
+- Uses: `sh -c "command"` under the hood
+
+### Direct Execution (Array)
+
+```ruby
+cmd(:my_cmd) { ["git", "status", "--porcelain"] }
+```
+
+- ✅ Safe from shell injection
+- ✅ Special characters treated as literals
+- ❌ No pipes, redirects, or variables
+- Uses: `Open3.popen3("git", "status", "--porcelain")` under the hood
+
+### With Working Directory
+
+```ruby
+config do
+  cmd(:in_tmp) { working_directory "/tmp" }
+end
+
+execute do
+  cmd(:in_tmp) { "pwd" }  # Runs in /tmp
+end
+```
+
+## Security Guidelines
+
+**Use shell commands (string) when:**
+
+- You control all input
+- You need shell features
+- Commands are static/trusted
+
+**Use direct execution (array) when:**
+
+- Handling user input
+- Special characters should be literal
+- Security is critical
+- You don't need shell features
+
+## Implementation Details
+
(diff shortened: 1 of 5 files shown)


mathiusj:
* **#523** <a href="https://app.graphite.dev/github/pr/Shopify/roast/523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/roast/523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/dsl/command_runner.rb
```diff
@@ -115,58 +95,102 @@ def execute(
...
             kill_process(pid)
-            wait_thread.join(1) # Give it a second to finish
+            wait_thread.join(1)
           end
         end
 
         private
 
+        #: (Integer, String, ?Integer?) -> void
+        def track_child_process(pid, command, pgid = nil)
```
nfgrep: Looks like this isn't called anywhere

lib/roast/dsl/command_runner.rb
```diff
@@ -115,58 +95,102 @@ def execute(
...
+          if pgid
+            kill_pgid(pgid)
+          else
+            kill_pid(pid)
           end
+        end
 
-          # If still running, use KILL signal
+        #: (Integer) -> void
+        def kill_process(pid)
```
nfgrep: Could just use `kill_pid` everywhere I think

PR #535, opened 2025-11-05
Parse Claude tool result messages

juniper-shopify:
Adds some rudimentary parsing for tool result messages. We need to store the tool use information in a context on the claude invocation, because the tool result only includes a `tool_use_id`


juniper-shopify (commented):
@dersam @nfgrep Sometimes the output of a tool result can be very long. we should decide on whether a sensible default is to truncate it or not (presumably with a configurable override either way)

dersam (approved):

juniper-shopify:
### Merge activity

* **Nov 5, 5:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/535).

PR #576, opened 2025-11-18
Replace Async {}.wait with Sync

juniper-shopify:
TIL `Sync` is a small optimization for a situation in which you want to run a block in an async context (so it has access to the event loop, can wait on other tasks, etc.), but the block itself is not going to be left in the background of the current context.

For example:
```
Async do
  ...
end.wait
```

is functionally identical to
```
Sync do
  ...
end
```

The main difference is that the `Sync` version will not create a new task if it is already inside a task, it will just execute it's block synchronously inside that asynchronous context.

### DIFF
--- a/lib/roast/dsl/command_runner.rb
+++ b/lib/roast/dsl/command_runner.rb
@@ -81,7 +81,7 @@ module Roast
           end
 
           # Read stdout and stderr concurrently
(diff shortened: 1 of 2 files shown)


nfgrep (approved):
Neat, TIL

mathiusj (approved):

PR #579, opened 2025-11-18
Use async barriers and annotate tasks

juniper-shopify:
This PR adds annotation to the workflow's async tasks, so when a task shows up in a stack trace or error output, it's easier to figure out what task had the problem.

Additionally, using an `Async::Barrier` to wait on the output of a set of tasks collectively provides slightly cleaner semantics when things go sideways. For instance, both of these will block until all tasks complete:

```
tasks = []
tasks << taskA = Async { ... }
tasks << taskB = Async { ... }
tasks.map(&:wait)
```

and

```
barrier = Async::Barrier.new
taskA = barrier.async { ... }
taskB = barrier.async { ... }
barrier.wait
```

The first one will not see an exception raised by Task B until after Task A finishes, even if Task B raised the exception much earlier. The second one will re-raise the exception as soon as it happens.

### DIFF
--- a/lib/roast/dsl/cog.rb
+++ b/lib/roast/dsl/cog.rb
@@ -57,11 +57,12 @@ module Roast
         @config = self.class.config_class.new #: untyped
       end
 
-      #: (Cog::Config, CogInputContext, untyped, Integer) -> Async::Task
-      def run!(config, input_context, executor_scope_value, executor_scope_index)
+      #: (Async::Barrier, Cog::Config, CogInputContext, untyped, Integer) -> Async::Task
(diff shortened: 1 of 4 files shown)


nfgrep (approved):

mathiusj:
just curious, is there any risk of deadlocks here that haven't been accounted for, when adding these barriers? It seems like we'd be able to handle some edge cases with the exception handling and barrier.stop calls, but I'm wondering about circular waits between tasks that don't hit exceptions.

mathiusj (approved):

PR #582, opened 2025-11-18
Enable shorthand template syntax with workflow directory resolution

mathiusj:
## Summary

Enables shorthand template syntax in DSL workflows by adding a comprehensive `template()` method that searches multiple locations for template files.

**Key changes:**

- Added `template()` method to `CogInputManager` with 13-location search strategy
- Added workflow directory tracking to `WorkflowContext`
- Enabled shorthand syntax: `template("greeting")` instead of full paths
- Comprehensive test coverage with realistic scenarios

## Template Resolution Strategy

The `template()` method searches in this priority order:

1. Absolute path as-is (if absolute)
2-4. Workflow directory: `path`, `path.erb`, `path.md.erb`
5-7. Workflow directory `prompts/`: `prompts/path`, `prompts/path.erb`, `prompts/path.md.erb`
8-10. Current directory: `path`, `path.erb`, `path.md.erb`
11-13. Current directory `prompts/`: `prompts/path`, `prompts/path.erb`, `prompts/path.md.erb`

## Testing

```bash
# Run all tests
bundle exec rake

# Run specific template tests
bundle exec ruby -Itest test/roast/dsl/cog_input_context_test.rb
```

**Test scenarios covered:**

- Shorthand path resolution from workflow directory
- Fallback to current working directory
- Absolute path handling
- Error handling for missing templates
- Real instances (no mocks) for production-like testing

## Files Changed

- `lib/roast/dsl/cog_input_manager.rb` - Added template resolution method
- `lib/roast/dsl/workflow_context.rb` - Added workflow_dir tracking
- `lib/roast/dsl/workflow.rb` - Pass workflow directory to context
- `test/roast/dsl/cog_input_context_test.rb` - Comprehensive test suite
- `dsl/simple_chat.rb` - Enabled shorthand syntax example
- Removed defunct method from `cog_input_context.rb`
- Updated Sorbet signatures

All review comments addressed, tests passing.


mathiusj:
* **#582** <a href="https://app.graphite.com/github/pr/Shopify/roast/582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/dsl/cog_input_manager.rb
```diff
@@ -151,6 +154,26 @@ def kwargs
       def tmpdir
         Pathname.new(@workflow_context.tmpdir).realpath
       end
+
+      #: (String, ?Hash) -> String
+      def template(path, args = {})
```
juniper-shopify: you should actually remove this method definition from `cog_input_context` when you add it here. it's getting overwritten by the binding, so it *works*, but it will be very confusing to have a defunct implementation lying around

lib/roast/dsl/cog_input_manager.rb
```diff
@@ -17,11 +17,12 @@ class CogFailedError < CogOutputAccessError; end
...
       class CogStoppedError < CogOutputAccessError; end
 
-      #: (Cog::Registry, Cog::Store, WorkflowContext) -> void
-      def initialize(cog_registry, cogs, workflow_context)
+      #: (Cog::Registry, Cog::Store, WorkflowContext, ?Pathname?) -> void
+      def initialize(cog_registry, cogs, workflow_context, workflow_dir = nil)
         @cog_registry = cog_registry
         @cogs = cogs
         @workflow_context = workflow_context
+        @workflow_dir = workflow_dir
```
juniper-shopify: add the workflow dir to the WorkflowContext, instead of passing it through explicitly. it should also be non-nilable

test/roast/dsl/cog_input_context_test.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      def teardown
+        FileUtils.rm_rf(@temp_dir) if @temp_dir && File.exist?(@temp_dir)
+      end
+
+      test "template method resolves shorthand paths relative to workflow directory when provided" do
+        workflow_dir = Pathname.new(@workflow_dir)
+        manager = CogInputManager.new(@cog_registry, @cogs, @workflow_context, workflow_dir)
+
+        # Change to different directory to test the fix
```
juniper-shopify: nit: test comments should not refer to "fixes". they should simply describe the objective behaviour being tested

test/roast/dsl/cog_input_context_test.rb
```diff
@@ -0,0 +1,92 @@
...
+          result = manager.context.template("test_template", { name: "Fallback" })
+          assert_includes result, "Hello Fallback!"
+          assert_includes result, "This is a test template."
+        end
+      end
+
+      test "template method fails with shorthand syntax when not in correct directory and no workflow_dir" do
+        manager = CogInputManager.new(@cog_registry, @cogs, @workflow_context, nil)
+
+        # This test demonstrates the bug when workflow_dir is not provided
```
juniper-shopify: nit: same as above -- test comment shouldn't refer to a bug that no longer exists, just describe the expected behaviour it is testing.

(okay, to be fair, there are cases where I think it's useful to have some comments in a test case explaining the particulars of a bug. But, I'd reserve that explicitly for when the bug is very complicated and counterintuitive, and non-obvious to understand what could go wrong to someone with even medium familiarity with the component)

juniper-shopify (changes requested):
thanks for doing this!

A more robust priority list of search locations will be helpful, and adding the workflow dir to the WorkflowContext will reduce the size of these method signatures that are already kind of too big

test/roast/dsl/cog_input_context_test.rb
```diff
@@ -0,0 +1,97 @@
...
+        # Change to temp_dir which has no prompts directory
+        Dir.chdir(@temp_dir) do
+          error = assert_raises(CogInputContext::ContextNotFoundError) do
+            manager.context.template("test_template", { name: "World" })
+          end
+          assert_includes error.message, "The file 'test_template' could not be found"
+        end
+      end
+
+      test "template method works with full path" do
```
juniper-shopify: are you intending to explicitly test an absolute path here (i.e. 'full path')? you're getting that kind of by accident because `Dir.mktmpdir` in the setup block is returning an absolute path, but it is not necessarily always the case that it will do so, if `Dir.tmpdir` is not itself an absolute path.

Call `@template_path.expand_path` here to get something that is for sure absolute

PR #613, opened 2025-11-24
Switch `print_*` methods to `show_*` in `cmd` cog config

juniper-shopify:


lib/roast/dsl/cogs/cmd.rb
```diff
@@ -5,87 +5,165 @@ module Roast
...
+          # - `show_stdout?`
+          # - `display!`
           #
           #: () -> void
-          def print_all!
-            # comment about how this method works
-            @values[:print_stdout] = true
-            @values[:print_stderr] = true
+          def show_stdout!
+            raise "⚠️ DEPRECATION: use #{__callee__.to_s.sub("print_", "show_")} instead of #{__callee__}" if __callee__.to_s.include?("print_")
```
mathiusj: possible or worth having functional tests for these raises?

juniper-shopify: I didn't think it was worth it.

I only put them in at all because I know that @dersam at least has already written some workflows that use these config values, and in case anyone else has I want to give them at least an explanation of why their workflow isn't running anymore.

This is all pre-pre-pre-release, so I'm not feeling beholden to any sort of backcompat or deprecation schedule, but I also don't want to be mean. I plan to just delete these in a few weeks when anyone who is actively using the incubating DSL will have noticed.

PR #631, opened 2025-11-28
Make JSON parsing a little safer

juniper-shopify:


lib/roast/dsl/cog/output.rb
```diff
@@ -104,6 +104,8 @@ def extract_code_blocks(input, language)
             end
 
             blocks
+          rescue
+            []
```
mathiusj: nothing relevant to raise or log here by chance? :D

juniper-shopify: Not to raise, since the purpose here is to be best effort and safe in the face of malformed input.

I could see the value of trace logging at some point. I think this is probably below the level of even the a debug log

PR #635, opened 2025-12-01
Add number parsing helpers for output types that return text

juniper-shopify:


lib/roast/dsl/cog/output.rb
```diff
@@ -131,6 +131,136 @@ def extract_json_like_blocks(input)
...
+
+            # 2. Try each line from bottom up (with whitespace stripped)
+            lines = input.lines.map(&:strip).reject(&:empty?)
+            candidates.concat(lines.reverse)
+
+            # 3. Try to extract numbers from each line from bottom up
+            lines.reverse.each do |line|
+              # Look for numbers with various separators, formats, and currency symbols (very permissive)
+              # Matches: 123, 1,234, 1_234, 1 234, 1.23, -1.23, 1.23e10, 1.23e-10
+              matches = line.scan(/-?[\d\s$¢£€¥.,_]+(?:[eE][+-]?\d+)?/) #: as Array[String]
```
mathiusj: should we store this regex as a global var or var somewhere with a useful name? could prevent the need for comment above on what this is looking for :)

mathiusj: same with the regex below actually

juniper-shopify:
### Merge activity

* **Dec 4, 8:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/635).

PR #642, opened 2025-12-01
Use OPENAI_API_BASE as openai api base url env var

juniper-shopify:
This seems to be the most common env var used for setting the OpenAI API base URL across various open-source project, although there really doesn't seem to be an official standard


mathiusj (approved):

mathiusj (approved):

juniper-shopify:
### Merge activity

* **Dec 4, 8:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/642).

PR #654, opened 2025-12-09
Fix sorbet issue in minitest_coverage_runner

juniper-shopify:
I was getting a weird Sorbet problem in a CI environment in a project that used Roast, caused by a collision between this instance of aliasing the `:original_at_exit` method and something else that was being done inside the minitest `.rbi` shim. This line is not used at all, and will be removed, so just changing the name for now seems to have resolved the problem


juniper-shopify:
### Merge activity

* **Dec 10, 10:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/654).

PR #655, opened 2025-12-09
Add VCR infrastructure for DSL functional tests

mathiusj:
# Add VCR Infrastructure for DSL Workflow Tests

Added VCR support for DSL workflow tests to enable reliable HTTP request recording and playback.

## What changed?

- Created a new `VCRURLRewriter` module to handle API endpoint transformations for cassette portability
- Added a test workflow `vcr_test_workflow.rb` that makes a simple LLM query about the deepest lake
- Created a VCR cassette with a recorded response about Lake Baikal
- Added a new test case that verifies the workflow runs correctly with the VCR cassette
- Set up environment variables in test setup to ensure tests work in both recording and replay modes

## How to test?

In replay mode (default):
```
bundle exec rake test
```

To record new cassettes:
```
OPENAI_API_KEY=your-key RECORD_VCR=true bundle exec rake test
```

## Why make this change?

This change improves test reliability by eliminating dependencies on external API services during testing. By recording API interactions once and playing them back during subsequent test runs, we can:

- Make tests faster and more deterministic
- Avoid hitting rate limits or incurring costs from real API calls
- Test without requiring real API credentials
- Ensure consistent test behavior regardless of network conditions

### DIFF
--- a/test/examples/support/functional_test.rb
+++ b/test/examples/support/functional_test.rb
@@ -15,10 +15,22 @@ module Examples
       FileUtils.mkdir_p(tmpdir_root) unless Dir.exist?(tmpdir_root)
 
       out, err = capture_io do
-        in_tmpdir(workflow_id.to_s, tmpdir_root) do |workflow_dir|
-          tmpdir = workflow_dir
-          FileUtils.cp_r(examples_source_path, workflow_dir)
-          block.call
+        # Set up test environment variables for VCR playback
+        # When recording (RECORD_VCR=true), uses real credentials from environment
+        # When playing back, uses fake credentials (VCR intercepts all requests)
+        recording = ENV["RECORD_VCR"]
+
+        with_env("OPENAI_API_KEY", recording ? ENV["OPENAI_API_KEY"] : "my-token") do
+          with_env("OPENAI_API_BASE", recording ? ENV["OPENAI_API_BASE"] : "http://mytestingproxy.local/v1") do
+            in_tmpdir(workflow_id.to_s, tmpdir_root) do |workflow_dir|
+              tmpdir = workflow_dir
+              FileUtils.cp_r(examples_source_path, workflow_dir)
+
+              VCR.use_cassette(workflow_id.to_s, record: recording ? :all : :none) do
+                block.call
+              end
+            end
+          end
         end
       end
 
--- a/test/fixtures/vcr_cassettes/simple_chat.yml
+++ b/test/fixtures/vcr_cassettes/simple_chat.yml
@@ -1,0 +1,10220 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: http://mytestingproxy.local/v1/chat/completions
+    body:
+      encoding: UTF-8
+      string: '{"model":"gpt-4o","messages":[{"role":"user","content":"What is the
+        deepest lake?"}],"stream":false}'
+    headers:
+      User-Agent:
+      - Faraday v2.13.1
+      Authorization:
+      - my-token
+      Content-Type:
+      - application/json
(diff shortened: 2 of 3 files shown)


test/dsl/functional/roast_dsl_examples_test.rb
```diff
@@ -338,6 +338,39 @@ class RoastDSLExamplesTest < FunctionalTest
...
+          stdout, stderr = in_sandbox :vcr_test do
+            File.write("dsl/vcr_test.rb", workflow_code)
+            result = Roast::DSL::Workflow.from_file("dsl/vcr_test.rb", EMPTY_PARAMS)
+            File.delete("dsl/vcr_test.rb")
+            result
+          end
+
+          assert_empty stderr
+          assert_predicate stdout, :present?
+          assert_match(/deepest/i, stdout)
```
juniper-shopify: This is not actually testing that the VCR is doing anything, just that the workflow is printing the word 'deepest' at some point

test/dsl/support/functional_test.rb
```diff
@@ -4,12 +4,27 @@ module DSL
...
     # Parameter workflow_id is an arbitrary namespace/subdirectory within the sandbox
     # Returns an array of strings [stdio_output, stderr_output]
     def in_sandbox(workflow_id, &block)
+      _run_in_sandbox(workflow_id, &block)
+    end
+
+    private
+
+    # Internal helper that sets up sandbox environment
+    def _run_in_sandbox(workflow_id, &block)
```
juniper-shopify: why do we need the internal `_run_in_sandbox` method? `in_sandbox` isn't doing anything else, and `_run_in_sandbox` is only called once here

test/fixtures/vcr_cassettes/dsl_vcr_test.yml
```diff
@@ -0,0 +1,441 @@
...
+      X-Ratelimit-Reset-Requests:
+      - 6ms
+      X-Ratelimit-Reset-Tokens:
+      - 0s
+      X-Request-Id:
+      - req_779f54594b3a4dce8a8103263e8d964a
+      Set-Cookie:
+      - __cf_bm=NVCz9veOxuxKvbDejQK.PLa.ozdGRLgBBHEaYFQ2nyM-1765324192.5862405-1.0.1.1-y_SR21CuQaXmAoBgAAtKeiOMrdD..F6mWcxcxNvNOLxuMUMLkkankEnJ.UoS74KZEnPImwYFfr3mj5jlDu7OcCRaQd7oUSuEkgHGqmEx9Z2a3.lETqpO0_lBtD2anq61;
+        HttpOnly; Secure; Path=/; Domain=proxy.shopify.ai; Expires=Wed, 10 Dec 2025
+        00:19:55 GMT
```
juniper-shopify: Don't include shopify's internal API proxy, or any secure cookies, in your recorded cassettes!

juniper-shopify (dismissed):
You're on the right track here. I think this is good.

A few functional issues to work through, but I like the direction

mathiusj:
> This doesn't actually work. You need to also delete the cassette file that you want to re-record.

@juniper-shopify I missed this comment earlier as I was focused on the inline comments, so would have to backtrack a bit to see what the issue was that you were getting. Mind sharing the stacktrace or response you were getting so I can see what the issue was? I'm also adding `OPENAI_API_BASE=...`  to run the test with it pointing to our proxy, which does work with the rest of that line. 

Please LMK if it works for you now! Also added a new class for rewriting the vcr config, with tests.  Given it's EOD EDT I imagine this one may carry over to the new year, so can split this work into a stack for better reviewing/testing, when I get back

PR #677, opened 2026-01-14
Simplify CI config to only run with Ruby 3.4 and ActiveSupport 8.0

juniper-shopify:
Roast is a tool (like Foreman), not a gem that you would generally include in a project. We can be stricter about the versions that we require

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,23 +14,14 @@ env:
   CI: true
 
 jobs:
-  rubies:
-    strategy:
-      fail-fast: false
-      matrix:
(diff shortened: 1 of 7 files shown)


mathiusj (approved):

juniper-shopify:
### Merge activity

* **Jan 16, 4:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/677).
* **Jan 16, 4:26 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/677).

PR #783, opened 2026-02-12
Refactor agent provider to allow support for custom agents

dersam:
This will allow us to support agents other than Claude Code, though that's not implemented yet.

Primarily, what this does is move provider definitions to a registry on the workflow, rather than a hardcoded array. Roast will continue to define a select set of default providers  such as Claude Code (and opencode, soon). Third parties can add their own definitions in gems, and these providers will be pulled in on the necessary workflows via `use` statements (not implemented yet, coming next).

### DIFF
--- a/lib/roast/cogs/agent/config.rb
+++ b/lib/roast/cogs/agent/config.rb
@@ -5,7 +5,17 @@ module Roast
   module Cogs
     class Agent < Cog
       class Config < Cog::Config
-        VALID_PROVIDERS = [:claude].freeze #: Array[Symbol]
+        #: (?Hash[Symbol, untyped]) -> void
+        def initialize(initial = {})
+          super(initial)
+
+          @provider_registry = {}
+        end
+
+        def validate!
+          # Provider registry is responsible for ensuring the validity of providers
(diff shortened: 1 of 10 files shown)


lib/roast/cogs/agent/config.rb
```diff
@@ -36,27 +46,6 @@ def use_default_provider!
...
-        # It will __not__, however, validate that the agent is properly installed on your system.
-        # If the agent is not properly installed, you will likely experience a failure when Roast attempts to
-        # run your workflow.
-        #
-        # #### See Also
-        # - `provider`
-        # - `use_default_provider!`
-        #
-        #: () -> Symbol
-        def valid_provider!
```
dersam: This validation is done by the registry now.

juniper-shopify (changes requested):
I like the overall approach, but I don't love the way it integrates with the config object. See my suggestions in line about including the provider registry on the workflow context object (or adding a new "cog resource context" object in parallel to it)

dersam:
Going to close this and reattempt in a separate stack, easier than trying to refactor out of this current state.

PR #784, opened 2026-02-12
Add example for defining new agent provider

dersam:
This is the most minimal possible agent. This one always responds with
the same thing and always succeeds. This should make it easier to
understand how to implement providers, as the Claude version is quite
complex.

Also we decided we don't care about duplicated agent responses in Claude
and removed some special casing.

### DIFF
--- a/tutorial/10_custom_cogs/README.md
+++ b/tutorial/10_custom_cogs/README.md
@@ -1,0 +1,455 @@
+![roast-horiz-logo](https://github.com/user-attachments/assets/f9b1ace2-5478-4f4a-ac8e-5945ed75c5b4)
+
+# Chapter 10: Custom Cogs
+
+🔥 _version 1.0 feature preview_ 🔥
+
+## Overview
+
+Roast's built-in cogs (`chat`, `agent`, `cmd`, `ruby`, `map`, `repeat`, `call`) cover many common use cases, but sometimes you need specialized functionality. Custom cogs let you:
+
+- **Integrate external services** - API clients, databases, cloud services
+- **Reuse complex logic** - Multi-step operations that you use across workflows
+- **Add domain-specific operations** - Business logic specific to your organization
+- **Create testable components** - Isolated units with clear inputs and outputs
+- **Package and share functionality** - As gems or local files
+
+This chapter teaches you how to create custom cogs that seamlessly integrate with Roast workflows.
+
+## When to Create a Custom Cog
+
+Create custom cogs when you need to:
+
+✅ **Wrap an external API or service** that you call frequently
+✅ **Encapsulate multi-step logic** that appears in multiple workflows
+✅ **Add organization-specific operations** (e.g., deploy to your infrastructure)
+✅ **Create reusable, testable components** with clear interfaces
+
+❌ Don't create a custom cog for one-off operations - use the `ruby` cog instead
+❌ Don't create a cog just to wrap a single method call - keep it simple
+
+## Anatomy of a Cog
+
+Every cog has three main components:
+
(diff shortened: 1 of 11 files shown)


tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+```ruby
+class MyCog < Roast::Cog
+  # 1. Config - Configuration options (optional)
+  class Config < Roast::Cog::Config
+    attr_accessor :timeout
+
+    def initialize
+      @timeout = 30
+    end
```
juniper-shopify: This won't work with config merging. With our architecture, config values *must* be stored in the `@values` hash.

```suggestion
    def timeout(seconds)
      @values[:timeout] = seconds
    end
    
    # by convention, we use `valid_<attibute>` method as accessors used by the cog, to distinguish them from setters used in the workflow definition
    def valid_timeout
      @values.fetch(:timeout, 30)
    end
```

tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+```ruby
+config do
+  my_cog do
+    timeout 60
+    retries 5
+  end
+end
+```
+
+### Input (Required)
```
juniper-shopify: We *could* make the input class optional (maybe should, if cog's with no inputs are a thing).

tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+class Simple < Roast::Cog
+  class Input < Roast::Cog::Input
+    def validate!
+      true
+    end
+  end
+
+  def execute(input)
+    puts "I'm a cog!"
+  end
```
juniper-shopify:
```suggestion
  def execute(input)
    puts "I'm a cog!"
    Cog::Output.new
  end
```

Right now `execute` must return an `Cog::Output` instance. As noted above, we could relax this

tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+execute do
+  weather do
+    city "San Francisco"
+  end
+
+  chat do
+    <<~PROMPT
+      The weather in #{weather!.city} is #{weather!.temperature}°F and #{weather!.conditions}.
+      Should I bring an umbrella?
+    PROMPT
```
juniper-shopify:
```suggestion
    city, temperature, conditions = weather!(:my_weather).values_at(:city, :temperature, :conditions)
    <<~PROMPT
      The weather in #{city} is #{temperature}°F and #{conditions}.
      Should I bring an umbrella?
    PROMPT
```

tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+For better organization, you can namespace your cogs:
+
+```ruby
+# lib/other.rb
+module MyCogNamespace
+  class Other < Roast::Cog
+    class Input < Roast::Cog::Input
+      def validate!
+        true
+      end
```
juniper-shopify:
```suggestion
      def validate!
      end
```

tutorial/10_custom_cogs/README.md
```diff
@@ -0,0 +1,455 @@
...
+    end
+
+    def execute(input)
+      puts "I'm a namespaced cog!"
```
juniper-shopify:
```suggestion
      puts "I'm a namespaced cog!"
      Cog::Output.new
```

juniper-shopify (changes requested):

PR #802, opened 2026-02-25
Introduce Event Framework

juniper-shopify:
This PR introduces an asynchronous event framework for Roast, and uses it to wrap the existing logging infrastructure. Subsequent PRs will build on this functionality to produce:
- a more robust and comprehensive user experience at the console, especially when running in an interactive terminal,
- better attribution of log messages and exception stack traces to the point in the cog execution hierarchy they occurred, and
- cleaner handling of the parallel output of asynchronously running cogs

The design principle here is that:
- `TaskContext` maintains a fiber-local stack that tracks where in the cog execution hierarchy a particular task is executing
- Anything that produces a log event or console output pushes `Event`s into a queue instead of emitting that output directly.
- `Event` automatically captures the current `TaskContext` state for the fiber in which the `Event` is produced.
- `EventMonitor` consumes the event queue in one fiber, allowing it to orchestrate how the output of multiple cogs running in parallel is produced. The goal here is to ultimately power a cleaner interactive UI that allows the user to see the state of multiple cogs executing in parallel in a clean and non-interleaved way, while also producing elegant log output for non-interactive sessions.

This PR does not change anything yet about the format of what Roast actually outputs. It simply adds the new framework. (It does add some rough new debug log statements for the beginning and ending of each cog's execution).

### DIFF
--- a/lib/roast/output_router.rb
+++ b/lib/roast/output_router.rb
@@ -1,0 +1,76 @@
+# typed: true
+# frozen_string_literal: true
+
+module Roast
+  class OutputRouter
+    # This is the name of the alias methods for `:write` on the $stdout and $stderr objects
+    # that bypass OutputRouter's wrapper. Calling this method will write to those output stream directly.
+    WRITE_WITHOUT_ROAST = :write_without_roast
+
+    class << self
+      #: () -> bool
+      def enable!
+        return false if enabled?
+
+        activate($stdout, :stdout)
+        activate($stderr, :stderr)
+        mark_as_output_fiber!
+        true
+      end
+
+      #: () -> bool
+      def disable!
+        return false unless enabled?
+
+        deactivate($stdout)
+        deactivate($stderr)
+        @output_fiber = nil
+        true
+      end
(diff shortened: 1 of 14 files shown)


lib/roast/output_router.rb
```diff
@@ -0,0 +1,74 @@
+# typed: true
+# frozen_string_literal: true
+
+module Roast
+  class OutputRouter
+    WRITE_WITHOUT_ROAST = :write_without_roast
```
dersam: Add a comment here - what is this meant for?

dersam (changes requested):

juniper-shopify:
### Merge activity

* **Feb 25, 10:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/802).
* **Feb 25, 10:26 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/802).

PR #803, opened 2026-02-25
Automatically capture logger output in all test cases

juniper-shopify:
Test cases should be silent unless they fail. This addition to `ActiveSupport::TestCase` captures the output of `Roast::Log` to an instance variable for every test case, and prints it to `stderr` if the test case fails.

(A test case can, of course, assert anything it wants on the logger output as well).

As we add more logging to Roast, the output of running various operations will become noiser. This is a good thing in general, but not for the test suite. I think the cleanest approach is a central extension like this, as opposed to adding a concern, or manual setup/teardown logic in each test class.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -30,6 +30,29 @@ if ENV["CI"]
   Minitest::RG.rg!(color: true)
 end
 
+# @requires_ancestor: ActiveSupport::TestCase
+module CaptureLogOutput
+  extend ActiveSupport::Concern
+
+  included do
+    setup do
+      @logger_output = StringIO.new
+      Roast::Log.logger = Logger.new(@logger_output)
+    end
+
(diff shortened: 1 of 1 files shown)


dersam (approved):

juniper-shopify:
### Merge activity

* **Feb 25, 10:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/803).
* **Feb 25, 10:28 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Shopify/roast/803) rebased this pull request as part of a merge.
* **Feb 25, 10:28 PM UTC**: @juniper-shopify merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/803).

PR #808, opened 2026-02-28
Add pastel gem for coloured terminal output

juniper-shopify:
I'm going to be adding a bunch of more nicely formatted terminal output when running roast in an interactive shell, alongside the work I'm doing to improve and augment the logging output. Adding support for colours is a first step.


lib/roast.rb
```diff
@@ -2,8 +2,8 @@
 # frozen_string_literal: true
 
 # Standard library requires
```
juniper-shopify: I alphabetized this list

PR #871, opened 2026-05-08
Support tilde (~) expansion in template paths

### LINKED ISSUE #663: Support tilde (~) expansion in template paths
## Problem

Currently, the `template()` method in DSL workflows uses `Pathname` which does not automatically expand `~` for the user's home directory.

While this is a niche use case, supporting it would allow workflows to reference templates in a user's home directory, which could be useful for shared personal configuration or reusable prompt libraries.

## Example

```ruby
# Currently doesn't work:
chat { template("~/my_prompts/greeting.md.erb", name: "World") }

# Would need to use:
chat { template(File.expand_path("~/my_prompts/greeting.md.erb"), name: "World") }
```

## Related

- PR #582 - template method with workflow directory resolution
- Review comment: https://github.com/Shopify/roast/pull/582#discussion_r1908346853

LasmarKhalifa:
Fixes https://github.com/Shopify/roast/issues/663

Paths with a tilde currently do not get expanded in the CogInputManager#template. This PR fixes that.

Adds File.expand_path candidates to the template path resolver so that ~/foo and ~user/foo resolve correctly.

Rescue block makes sure we don't crash when expanding ~something/foo (assuming "something" is not a real user) and fall back to the other candidates instead.

Added test coverage

### DIFF
--- a/lib/roast/cog_input_manager.rb
+++ b/lib/roast/cog_input_manager.rb
@@ -177,11 +177,10 @@ module Roast
     # 5-7. Workflow directory prompts/: prompts/path, prompts/path.erb, prompts/path.md.erb
     # 8-10. Current directory: path, path.erb, path.md.erb
     # 11-13. Current directory prompts/: prompts/path, prompts/path.erb, prompts/path.md.erb
+    # 14-16. Tilde-expanded path: path, path.erb, path.md.erb
     #
     #: (String | Pathname, ?Hash) -> String
     def template(path, args = {})
-      # NOTE: Pathname does not expand ~ for home directory automatically.
-      # This is tracked in issue https://github.com/Shopify/roast/issues/663.
       path = Pathname.new(path) unless path.is_a?(Pathname)
 
       # Priority stack of places to look for a matching file
@@ -212,6 +211,15 @@ module Roast
       candidate_paths << pwd / "prompts" / "#{path}.erb"
       candidate_paths << pwd / "prompts" / "#{path}.md.erb"
 
+      # 14-16. Tilde expanded path
+      begin
+        expanded_path = Pathname.new(File.expand_path(path))
+        candidate_paths << expanded_path
+        candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.erb"))
+        candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.md.erb"))
+      rescue ArgumentError
+        # skip
+      end
       # Use the first path that exists
       resolved_path = candidate_paths.find(&:exist?)
 
--- a/lib/roast/cog_input_manager.rb
+++ b/lib/roast/cog_input_manager.rb
@@ -218,7 +218,8 @@ module Roast
         candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.erb"))
         candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.md.erb"))
       rescue ArgumentError
-        # skip
+        # File.expand_path raises when expanding ~something/foo (assuming "something" is not a real user).
+        # Nothing to do here, falls back to other candidate paths without tilde expansion.
       end
       # Use the first path that exists
       resolved_path = candidate_paths.find(&:exist?)
--- a/test/roast/cog_input_manager_test.rb
+++ b/test/roast/cog_input_manager_test.rb
@@ -67,6 +67,41 @@ module Roast
       assert_includes result, "This is a test template."
     end
 
+    test "template method expands user home directories" do
+      Tempfile.create(["temp_file", ".md.erb"], File.expand_path("tmp")) do |file|
+        File.write(file, <<~ERB)
+          Hello <%= name %>!
+          This is a test template.
+        ERB
+
+        manager = create_manager
+
+        tilde_path = "~/#{Pathname.new(File.expand_path(file)).relative_path_from(Dir.home)}"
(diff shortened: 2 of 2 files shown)


LasmarKhalifa:
* **#872** <a href="https://app.graphite.com/github/pr/Shopify/roast/872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#871** <a href="https://app.graphite.com/github/pr/Shopify/roast/871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/871?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/cog_input_manager.rb
```diff
@@ -212,6 +211,15 @@ def template(path, args = {})
...
       candidate_paths << pwd / "prompts" / "#{path}.md.erb"
 
+      # 14-16. Tilde expanded path
+      begin
+        expanded_path = Pathname.new(File.expand_path(path))
+        candidate_paths << expanded_path
+        candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.erb"))
+        candidate_paths << Pathname.new(File.expand_path("#{expanded_path}.md.erb"))
+      rescue ArgumentError
+        # skip
```
juniper-shopify: I had just written 'skip' here temporarily while we were pairing, but best practice is generally to explain _why_ it's okay to swallow the exception here, so that future devs don't have to think to hard about it when making changes to this logic.

test/roast/cog_input_manager_test.rb
```diff
@@ -67,6 +67,41 @@ def teardown
       assert_includes result, "This is a test template."
     end
 
+    test "template method expands user home directories" do
+      Tempfile.create(["temp_file", ".md.erb"], File.expand_path("tmp")) do |file|
```
juniper-shopify: I'd prefer we name this temp file something more descriptive of its actual use here, like "test_template_in_user_home"

Ideally I'd like to create it in a subdirectory of `./tmp` as well, but that makes the automatic cleanup much messier and error prone, because we'd also want to clean up that subdirectory but only if it's empty, etc. so I think the best course of action given the tradeoffs is to put it directly in `./tmp`like you have here, but that's an extra reason why a more descriptive file name would be good. Mostly it doesn't matter that much since the file should get deleted after the test case completes, but in the off chance that it does not...

test/roast/cog_input_manager_test.rb
```diff
@@ -67,6 +67,41 @@ def teardown
...
+        File.write(file, <<~ERB)
+          Hello <%= name %>!
+          This is a test template.
+        ERB
+
+        manager = create_manager
+
+        tilde_path = "~/#{Pathname.new(File.expand_path(file)).relative_path_from(Dir.home)}"
+        result = manager.context.template(tilde_path, { name: "Full Path" })
```
juniper-shopify: I'd use "Relative path from user home" or something like that instead of "Full Path" as the name here

test/roast/cog_input_manager_test.rb
```diff
@@ -67,6 +67,41 @@ def teardown
...
+          tilde_path = "~something/template.md.erb"
+          FileUtils.mkdir_p("~something")
+          File.write(tilde_path, <<~ERB)
+            Hello <%= name%>!
+            This is a test template.
+          ERB
+
+          manager = create_manager
+
+          result = manager.context.template(tilde_path, { name: "Full Path" })
```
juniper-shopify: similarly, better to replace "Full Path" with something like "Path with literal tilde"

test/roast/cog_input_manager_test.rb
```diff
@@ -67,6 +67,41 @@ def teardown
...
+        manager = create_manager
+
+        tilde_path = "~/#{Pathname.new(File.expand_path(file)).relative_path_from(Dir.home)}"
+        result = manager.context.template(tilde_path, { name: "Full Path" })
+        assert_includes result, "Hello Full Path!"
+        assert_includes result, "This is a test template."
+      end
+    end
+
+    test "template method resolves file with literal tilde" do
```
juniper-shopify: This test case is actually testing "template method handles ~user reference to non-existent user gracefully" more than literal tilde.

This is a good thing to test, but we should also a test case with the literal tilde somewhere in the middle of the path, just to confirm that that works, and also a test case where the path given to template is something like `~/file_name.md.erb` but the user's intent is a subdirectory of the current directory that is named \*just* a single literal tilde

juniper-shopify (changes requested):
Looking very good! Just a couple tweaks and I think it'd be good to add a few more test cases for completeness

LasmarKhalifa:
> Looking very good! Just a couple tweaks and I think it'd be good to add a few more test cases for completeness

Thanks for the feedback!! Did the requested changes :rocket:

juniper-shopify (approved):
🔥

LasmarKhalifa:
### Merge activity

* **May 12, 8:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/871).
* **May 12, 8:31 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/871).

PR #893, opened 2026-05-21
Repeat log prefix on each line of multi-line output

### LINKED ISSUE #876: Repeat log prefix on each line of multi-line output
On multi-line log output, we currently only show the prefix once, causing the rest of the lines to wrap around it, which can make it confusing to see which continuation lines are associated with which cogs' output.

Example: 

Image

The goal is to make multi-line log outputs split on newlines and show the prefix again, making the output more consistent.

LasmarKhalifa:
fixes https://github.com/Shopify/roast/issues/876

The goal of this PR is to clean up the output of workflow runs by making it easier to read and understand which lines are part of which action.

## Examples

Old terminal output                                                                New terminal output:  
             

Old log output                                                                        New log output  
                     

Note: went with dots on continuation lines instead of whitespace so the "missing" prefix reads as intentional rather than looking like a rendering glitch or a dropped line. It also avoids touching `msg2str`, which strips leading/trailing whitespace. Any future `.strip` added elsewhere in the pipeline could also silently collapse a whitespace prefix.

### DIFF
--- a/lib/roast/log_formatter.rb
+++ b/lib/roast/log_formatter.rb
@@ -27,9 +27,9 @@ module Roast
 
     #: (String, String) -> String
     def colourize(severity, line)
-      if line.include?("❯❯") # standard error lines
+      if line.include?("❯❯") || line.include?("❙❙") # standard error lines
         @rainbow.wrap(line).yellow
-      elsif line.include?("❯") # standard output lines
+      elsif line.include?("❯") || line.include?("❙") # standard output lines
         @rainbow.wrap(line)
       else
         case severity
--- a/test/roast/event_monitor_test.rb
+++ b/test/roast/event_monitor_test.rb
@@ -188,14 +188,16 @@ module Roast
 
       event = Event.new([em_el, cog_el], { info: "line 1\nline 2\nline 3" })
       EventMonitor.accept(event)
-      lines = @logger_output.string.lines
-      prefix = "test_cog(:my_step)"
 
-      assert_equal 1, lines.count { |line| line.include?(prefix) }
-      assert_includes lines.find { |line| line.include?("line 1") }, prefix
-      refute_includes lines.find { |line| line.include?("line 2") }, prefix
-      refute_includes lines.find { |line| line.include?("line 3") }, prefix
-      assert_equal 2, lines.count { |line| line.include?("·") }
+      messages = @logger_output.string.gsub(/^.*? -- /, "")
+      prefix = "test_cog(:my_step)"
+      continuation_prefix = "·" * prefix.length
+
+      assert_equal(<<~OUTPUT, messages)
+        #{prefix} line 1
+        #{continuation_prefix} line 2
+        #{continuation_prefix} line 3
+      OUTPUT
     end
 
     test "handle_begin_event logs cog begin events" do
@@ -388,15 +390,16 @@ module Roast
 
       event = Event.new([em_el, cog_el], { stdout: "line 1\nline 2\nline 3" })
       EventMonitor.accept(event)
-      lines = @logger_output.string.lines
-      prefix = "test_cog(:my_step)"
 
-      assert_equal 1, lines.count { |line| line.include?(prefix) }
-      assert_includes lines.find { |line| line.include?("line 1") }, prefix
-      refute_includes lines.find { |line| line.include?("line 2") }, prefix
-      refute_includes lines.find { |line| line.include?("line 3") }, prefix
-      assert_equal 1, lines.count { |line| line.include?(" ❯ ") }
-      assert_equal 2, lines.count { |line| line.include?(" ❙ ") }
+      messages = @logger_output.string.gsub(/^.*? -- /, "")
+      prefix = "test_cog(:my_step)"
+      continuation_prefix = "·" * prefix.length
+
+      assert_equal(<<~OUTPUT, messages)
+        #{prefix} ❯ line 1
+        #{continuation_prefix} ❙ line 2
+        #{continuation_prefix} ❙ line 3
+      OUTPUT
     end
 
     test "handle_stderr_event does not output to console" do
@@ -422,15 +425,16 @@ module Roast
 
       event = Event.new([em_el, cog_el], { stderr: "line 1\nline 2\nline 3" })
       EventMonitor.accept(event)
-      lines = @logger_output.string.lines
-      prefix = "test_cog(:my_step)"
 
-      assert_equal 1, lines.count { |line| line.include?(prefix) }
-      assert_includes lines.find { |line| line.include?("line 1") }, prefix
-      refute_includes lines.find { |line| line.include?("line 2") }, prefix
-      refute_includes lines.find { |line| line.include?("line 3") }, prefix
-      assert_equal 1, lines.count { |line| line.include?(" ❯❯ ") }
-      assert_equal 2, lines.count { |line| line.include?(" ❙❙ ") }
+      messages = @logger_output.string.gsub(/^.*? -- /, "")
+      prefix = "test_cog(:my_step)"
+      continuation_prefix = "·" * prefix.length
+
+      assert_equal(<<~OUTPUT, messages)
+        #{prefix} ❯❯ line 1
+        #{continuation_prefix} ❙❙ line 2
+        #{continuation_prefix} ❙❙ line 3
+      OUTPUT
     end
 
     test "handle_unknown_event logs unrecognized events at unknown level" do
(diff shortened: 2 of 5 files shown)


LasmarKhalifa:
* **#902** <a href="https://app.graphite.com/github/pr/Shopify/roast/902?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#901** <a href="https://app.graphite.com/github/pr/Shopify/roast/901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#900** <a href="https://app.graphite.com/github/pr/Shopify/roast/900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#899** <a href="https://app.graphite.com/github/pr/Shopify/roast/899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#898** <a href="https://app.graphite.com/github/pr/Shopify/roast/898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#893** <a href="https://app.graphite.com/github/pr/Shopify/roast/893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

juniper-shopify:
> Note: went with dots on continuation lines instead of whitespace so the "missing" prefix reads as intentional rather than looking like a rendering glitch or a dropped line. It also avoids touching msg2str, which strips leading/trailing whitespace. Any future .strip added elsewhere in the pipeline could also silently collapse a whitespace prefix.

I like that idea. The dots do look intentional, and it is more robust to potential whitespace stripping :)

lib/roast/log_formatter.rb
```diff
@@ -27,9 +27,9 @@ def call(severity, time, _progname, msg)
 
     #: (String, String) -> String
     def colourize(severity, line)
-      if line.include?("❯❯") # standard error lines
+      if line.include?("❯❯") || line.include?("❙❙") # standard error lines
```
juniper-shopify, line 30: This was already a little hacky, matching the `❯` character in the line. I was kind of okay with it when I wrote it because it's an atypical character and the risk of false positives was low, although I'd wanted something better and more explicit. The `|` character is a very common character and we're going to get a lot of false positives with this.

We'll need to come up with a first-class way of indicating that a line should be coloured or not

test/roast/event_monitor_test.rb
```diff
@@ -181,6 +181,23 @@ class EventMonitorTest < ActiveSupport::TestCase
...
+      event = Event.new([em_el, cog_el], { info: "line 1\nline 2\nline 3" })
+      EventMonitor.accept(event)
+      lines = @logger_output.string.lines
+      prefix = "test_cog(:my_step)"
+
+      assert_equal 1, lines.count { |line| line.include?(prefix) }
+      assert_includes lines.find { |line| line.include?("line 1") }, prefix
+      refute_includes lines.find { |line| line.include?("line 2") }, prefix
+      refute_includes lines.find { |line| line.include?("line 3") }, prefix
+      assert_equal 2, lines.count { |line| line.include?("·") }
```
juniper-shopify: I think it would be more robust if this asserted the correct number of `.` characters in the expected continuation prefix

juniper-shopify: Even clearer might be to just assert that the full logger output equals an expected multiline string exactly (you'll need to strip out timestamps, of course). that's both comprehensive and really easy to read and reason about when you're unfamiliar with this test case and have just made a seemingly unrelated change that caused it to fail. This implementation takes a little thinking to parse.

test/roast/event_monitor_test.rb
```diff
@@ -364,6 +381,24 @@ class EventMonitorTest < ActiveSupport::TestCase
...
+      EventMonitor.accept(event)
+      lines = @logger_output.string.lines
+      prefix = "test_cog(:my_step)"
+
+      assert_equal 1, lines.count { |line| line.include?(prefix) }
+      assert_includes lines.find { |line| line.include?("line 1") }, prefix
+      refute_includes lines.find { |line| line.include?("line 2") }, prefix
+      refute_includes lines.find { |line| line.include?("line 3") }, prefix
+      assert_equal 1, lines.count { |line| line.include?(" ❯ ") }
+      assert_equal 2, lines.count { |line| line.include?(" ❙ ") }
```
juniper-shopify: same comments as above

test/roast/event_monitor_test.rb
```diff
@@ -380,6 +415,24 @@ class EventMonitorTest < ActiveSupport::TestCase
...
+      lines = @logger_output.string.lines
+      prefix = "test_cog(:my_step)"
+
+      assert_equal 1, lines.count { |line| line.include?(prefix) }
+      assert_includes lines.find { |line| line.include?("line 1") }, prefix
+      refute_includes lines.find { |line| line.include?("line 2") }, prefix
+      refute_includes lines.find { |line| line.include?("line 3") }, prefix
+      assert_equal 1, lines.count { |line| line.include?(" ❯❯ ") }
+      assert_equal 2, lines.count { |line| line.include?(" ❙❙ ") }
+    end
```
juniper-shopify, line 438: same comments as above

test/roast/log_formatter_test.rb
```diff
@@ -199,6 +199,18 @@ class LogFormatterTest < ActiveSupport::TestCase
       assert_includes result, "❯❯"
     end
 
+    test "stderr continuation lines are yellow in TTY mode" do
+      formatter = LogFormatter.new(tty: true)
+      time = Time.new(2026, 1, 1)
+
+      result = formatter.call("INFO", time, nil, "···· ❙❙ other line of some error output")
```
juniper-shopify, line 206: Writing the test case like this assumes half of the implementation you're testing, and only tests the other half.

What if we changed something so that continuation lines don't have `||` in them anymore, but we also changed the colourizer so they still got coloured correctly?

juniper-shopify: Of course, I'm also recommending that we don't trigger the formatter on `||`, so this would have to change anyway

test/roast/log_formatter_test.rb
```diff
@@ -210,6 +222,16 @@ class LogFormatterTest < ActiveSupport::TestCase
       assert_includes result, "❯"
     end
 
+    test "stdout continuation lines are not colourized in TTY mode" do
+      formatter = LogFormatter.new(tty: true)
+      time = Time.new(2026, 1, 1)
+
+      result = formatter.call("INFO", time, nil, "···· ❙ other line of some output")
```
juniper-shopify, line 229: Same comment as above

juniper-shopify (changes requested):

lib/roast/log_formatter.rb
LasmarKhalifa: It's actually `❙`, instead of`|`. I put some`|` characters to show that it doesn't get formatted the same way (in the examples above). In this case, would it still be a blocker for this PR?

test/roast/event_monitor_test.rb
LasmarKhalifa: Done!

test/roast/event_monitor_test.rb
LasmarKhalifa: Done!

test/roast/event_monitor_test.rb
LasmarKhalifa: Done!

lib/roast/log_formatter.rb
juniper-shopify: Oh I see. Yeah, that does mitigate the risk somewhat. Nice thought!

I'm still not crazy about keeping this pattern at all, though. Now that we are touching it is a good chance to improve it. I saw you put up a draft PR with a more robust implementation. I like that and I think we should go that way. But, I also think it's fine to merge this PR independently of that other one. Given the special unicode pipe-like character you're using, it's not really making the problem any worse

juniper-shopify (approved):
I'm happy with this as it is, as long as we follow up with a better version of stdout/stderr detection for the colourizer, like you're already working on in https://github.com/Shopify/roast/pull/903

LasmarKhalifa:
### Merge activity

* **May 28, 7:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/893).
* **May 28, 7:49 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/893).

Issue #896, opened 2026-05-26
[DISCUSSION] Potential approaches to prompt and response formatting


LasmarKhalifa:
This thread is to discuss potential approaches to having better prompt and response formatting.

LasmarKhalifa:
One way to do this would be to make chat/agent cogs' agent and responses appear as a "block of data" instead of the way we are currently showing it. This is useful because a long block of text now becomes clearly formatted with the prefix on a different line, removing the possibility of wrapping around it which makes the output much clearer and easier to reason about.

Current way: 
<img width="2282" height="146" alt="Image" src="https://github.com/user-attachments/assets/ba1d70e2-2d21-4842-ad77-ed71b6ef388b" />

Proposed way: 
<img width="2284" height="314" alt="Image" src="https://github.com/user-attachments/assets/430e0a73-cbcb-47fa-b48d-6bc0f25237e4" />

We could implement a new event type that passes a header and content, and then have an event handler in EventMonitor that formats it in a similar way as shown in the image above.

Prototype: https://app.graphite.com/github/pr/Shopify/roast/897

juniper-shopify:
> One way to do this would be to make chat/agent cogs' agent and responses appear as a "block of data"

Worth adding a note about why/how this helps with visual appearance despite not truncating any individual long lines (as we decided not to do in https://github.com/Shopify/roast/issues/877)

dersam:
I like this approach, let's do it.

LasmarKhalifa:
> > One way to do this would be to make chat/agent cogs' agent and responses appear as a "block of data"
> 
> Worth adding a note about why/how this helps with visual appearance despite not truncating any individual long lines (as we decided not to do in [#877](https://github.com/Shopify/roast/issues/877))

Added it!

LasmarKhalifa:
> I like this approach, let's do it.

Amazing!! Thanks for the input 🙌

LasmarKhalifa:
Closing this as we have reached a decision we are all aligned on.

PR #897, opened 2026-05-26
[PROTOTYPE] Improve formatting of agent prompt and response outputs

LasmarKhalifa:
Part of the exploration for https://github.com/Shopify/roast/issues/896

Prototype for creating a new event type for long multi-line messages and applying it to prompts and response messages for agent cogs.

This is just to demonstrate the functionality and proposed implementation.

Old formatting:  

New formatting:

### DIFF
--- a/lib/roast/cogs/agent/providers/claude/claude_invocation.rb
+++ b/lib/roast/cogs/agent/providers/claude/claude_invocation.rb
@@ -77,7 +77,7 @@ module Roast
               raise ClaudeAlreadyStartedError if started?
 
               @started = true
-              puts "[USER PROMPT] #{@prompt}" if @show_prompt
+              Event << { block: { header: "USER PROMPT", content: @prompt } } if @show_prompt
               _stdout, stderr, status = CommandRunner.execute(
                 command_line,
                 working_directory: @working_directory,
@@ -87,7 +87,7 @@ module Roast
 
               if status.success?
                 @completed = true
-                puts "[AGENT RESPONSE] #{@result.response}" if @show_response
+                Event << { block: { header: "AGENT RESPONSE", content: @result.response } } if @show_response
               else
                 @failed = true
                 @result.success = false
--- a/lib/roast/event_monitor.rb
+++ b/lib/roast/event_monitor.rb
@@ -130,6 +130,15 @@ module Roast
     end
 
     #: (Event) -> void
+    def handle_block_event(event)
+      block = event[:block]
+      header = "[#{block[:header]}]"
+      content = block[:content]
+      separator = "-" * 40
+      Roast::Log.logger.info { "#{format_path(event)} ↓\n\n#{header}\n#{separator}\n#{content}\n#{separator}" }
+    end
+
+    #: (Event) -> void
     def handle_unknown_event(event)
       Roast::Log.logger.unknown(event.inspect)
     end
--- a/lib/roast/event.rb
+++ b/lib/roast/event.rb
@@ -24,6 +24,7 @@ module Roast
       :end,
       :stdout,
       :stderr,
+      :block,
     ].freeze #: Array[Symbol]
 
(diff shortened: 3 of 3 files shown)


LasmarKhalifa:
* **#897** <a href="https://app.graphite.com/github/pr/Shopify/roast/897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/event_monitor.rb
```diff
@@ -129,6 +129,15 @@ def handle_stdout_event(event)
       Roast::Log.logger.info { "#{format_path(event)} ❯ #{event[:stdout]}" }
     end
 
+    #: (Event) -> void
+    def handle_block_event(event)
+      block = event[:block]
+      header = "[#{block[:header]}]"
+      content = block[:content]
+      separator = "-" * 40
+      Roast::Log.logger.info { "#{format_path(event)} ↓\n\n#{header}\n#{separator}\n#{content}\n#{separator}" }
```
juniper-shopify, line 138: I'd like to see what it looks like to put the header on the same line as the prefix, and then `{separator}\n{content}\n{separator}` below.

Also lets consider how it looks with `---` separators before and after, as you've done here, vs `vvv` before and `^^^` after separators

lib/roast/cogs/agent/providers/claude/claude_invocation.rb
```diff
@@ -77,7 +77,7 @@ def run!
               raise ClaudeAlreadyStartedError if started?
 
               @started = true
-              puts "[USER PROMPT] #{@prompt}" if @show_prompt
+              Event << { block: { header: "USER PROMPT", content: @prompt } } if @show_prompt
```
juniper-shopify, line 80: nice. this is clean and simple :)

probably worth extending the `block` event just a little, though, to encapsulate whether the block output is normal/expected (like stdout content) or abnormal/error information (like stderr content). This could be done with two block typed (`stdout_block` or `stderr_block`) or we could add a `severity` / `type` attribute to the block event's payload hash, etc. What do you think of the tradeoffs between these approaches?

LasmarKhalifa: My reasoning is that currently, the only two places we want to have create block event types are prompt and response messages from agent/chat cogs. Those two will not be displaying block error messages.

If we ever need an error block event, we would probably want it to have its own distinct formatting and its own handler (to have, for example, a different separator), in which case it would be better to have a separate event for it. It would also match the current pattern of stdout and stderr.

The tradeoff is that if they are very similar, every change would have to be done in both places. But if we use the type attribute now, we end up with 4 callers having to pass an extra parameter and none passing the error one, which is just extra unnecessary complexity.

I think the best approach is to have this one named as `block` and, if we eventually need an error block, create one called `error_block` with its own handler and formatting.

juniper-shopify: Yeah, I think that makes sense. A good example of not prematurely optimizing.

lib/roast/event_monitor.rb
juniper-shopify: Also, I think it'd be better to remove the extra blank like between the header and the block body, and also try putting header on the first line (with the prefix) instead of after it. I'm curious if that will look cleaner.

Which line the header goes on I feel less strongly about, but I do prefer not having the extra blank line. I think it makes the log output feel unnecessarily disconnected, and actually makes the body of the block look less associated with the header line, and more associated with whatever log line follows it

LasmarKhalifa: I actually tested a few approaches, and I agree that the blank space should be removed. I like number 2 on the list:

PR #903, opened 2026-05-28
Tag log messages with originating stream

### LINKED ISSUE #909: Refactor LogFormatter#colourize to be more robust
We currently use the ❙ and ❯ characters to determine whether we colourize an output a certain way or not. This should be refactored to be more robust and so that the associated tests become stronger as well (currently, if we change the character in the event handler and in the colourizer, the tests would fail although the behaviour would be good).

LasmarKhalifa:
closes https://github.com/Shopify/roast/issues/909

Refactors the log messages with a `type `attribute so that LogFormatter#colourize becomes more robust and tests assess the colourize method properly.

Adds .to_s to CustomLogging because we previously only had strings as messages, but we now have either a string or a Roast::Log::Message, so we need to normalize using .to_s so that we output only the text of the message rather than the entire object.

### DIFF
--- a/lib/roast/event_monitor.rb
+++ b/lib/roast/event_monitor.rb
@@ -130,7 +130,12 @@ module Roast
       path = format_path(event)
       event[:stderr].lines.each_with_index do |line, idx|
         path_prefix = idx.zero? ? "#{path} ❯❯" : "#{"·" * path.length} ❙❙"
-        Roast::Log.logger.warn { "#{path_prefix} #{line.rstrip}" }
+        Roast::Log.logger.warn do
+          Roast::Log::Message.new(
+            type: :stderr,
+            text: "#{path_prefix} #{line.rstrip}",
+          )
+        end
       end
     end
 
@@ -139,7 +144,12 @@ module Roast
       path = format_path(event)
       event[:stdout].lines.each_with_index do |line, idx|
         path_prefix = idx.zero? ? "#{path} ❯" : "#{"·" * path.length} ❙"
-        Roast::Log.logger.info { "#{path_prefix} #{line.rstrip}" }
+        Roast::Log.logger.info do
+          Roast::Log::Message.new(
+            type: :stdout,
+            text: "#{path_prefix} #{line.rstrip}",
+          )
+        end
       end
     end
 
--- a/lib/roast/log.rb
+++ b/lib/roast/log.rb
@@ -21,6 +21,27 @@ module Roast
     extend self
     include Kernel
 
+    # A log message paired with its originating type, so downstream handlers
+    # (e.g. the formatter) can render or route it by type.
+    class Message
+      #: Symbol
+      attr_reader :type
+
+      #: String
+      attr_reader :text
+
+      #: (type: Symbol, text: String) -> void
+      def initialize(type:, text:)
+        @type = type
+        @text = text
+      end
+
+      #: () -> String
+      def to_s
+        text
+      end
+    end
+
     LOG_LEVELS = {
       DEBUG: ::Logger::DEBUG,
       INFO: ::Logger::INFO,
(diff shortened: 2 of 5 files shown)


LasmarKhalifa:
* **#903** <a href="https://app.graphite.com/github/pr/Shopify/roast/903?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/903?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/event_monitor.rb
```diff
@@ -121,12 +121,18 @@ def handle_log_event(event)
 
     #: (Event) -> void
     def handle_stderr_event(event)
-      Roast::Log.logger.warn { "#{format_path(event)} ❯❯ #{event[:stderr]}" }
+      Roast::Log.logger.warn(Roast::Log::Message.new(
+        stream: :stderr,
+        text: "#{format_path(event)} ❯❯ #{event[:stderr]}",
+      ))
```
juniper-shopify: This is really clever. I like the approach of passing through a richer context object than just a string, so we can include additional parameters like the originating stream.

A couple minor optimizations I'd suggest:

The style of the existing logger invocation that might seem somewhat odd at first glance, where the message argument is passed inside a block passed to the `logger.warn` method instead of just passing the message directly as a string serves a useful purpose. The block is not evaluated at all if the relevant log level is disabled, saving the cost of the string formatting that happens inside it (and any additional work, which might in some cases be more substantial, like computing a full stack trace, etc.). I'd like to preserve that style of invocation, so please look into returning the `Roast::Log::Message` from a block instead of passing it directly (and because it now includes object instantiation, not just string formatting, there is actually a bit more perf overhead to save when the log won't actually be emitted).

I'd also suggest you consider renaming the `stream` parameter to something more generic, like `type`. Right now, special colouring rules are only used for stdout and stderr, but it might be the case that in the future we want additional special colours (for instance, we might want them for block output for prompts and/or responses of chat and agent cogs), so making that param a little more generic makes it more future-proof for this kind of thing

LasmarKhalifa: Done! The performance concern is definitely a good point. Thanks!!

lib/roast/log.rb
```diff
@@ -21,6 +21,27 @@ module Log
     extend self
     include Kernel
 
+    # Wraps a log message with the originating type so the formatter can
+    # colourize without parsing characters back out of the formatted line.
```
dersam: > without parsing characters back out of the formatted line.

What does that mean? What's the failure mode this is addressing - can you add an example?

LasmarKhalifa: We currently (in the LogFormatter#colourize) derive whether a message is stderr or stdout based on whether the string of the message contains `❙`,` ❯`, `❯❯` or `❙❙`

So this is addressing the current hacky implementation and making it more robust by providing the `type` of the message explicitly, so that it can be assessed without checking if a certain arbitrary character is in the message string or not (also makes our tests more robust). "parsing" is not the right word though, and after merging this we won't need to know what the old behaviour was, so I changed the comment to something more descriptive of the class itself.   
  
New comment:  
"A log message paired with its originating type, so downstream handlers (e.g. the formatter) can render or route it by type."  
  
Thanks for flagging this!!

dersam (approved):

LasmarKhalifa:
### Merge activity

* **Jun 3, 3:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/903).
* **Jun 3, 3:15 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/903).

PR #913, opened 2026-06-02
Replace case statement in Claude Message.from_hash with dynamic class resolution

### LINKED ISSUE #911: Replace case statement in Claude Message.from_hash with dynamic class resolution
## Summary

`Message.from_hash` in `lib/roast/cogs/agent/providers/claude/message.rb` uses an 8-branch `case` statement to map a `:type` symbol to a `Messages::*Message` class. Since every branch follows the exact same naming convention (`:"#{type}_message".to_s.camelize`), we can replace the static dispatch with dynamic constant resolution.

## Current code (lines 23–44)

```ruby
def from_hash(hash)
  type = hash.delete(:type)&.to_sym
  case type
  when :assistant
    Messages::AssistantMessage.new(type:, hash:)
  when :result
    Messages::ResultMessage.new(type:, hash:)
  when :system
    Messages::SystemMessage.new(type:, hash:)
  when :text
    Messages::TextMessage.new(type:, hash:)
  when :thinking
    Messages::ThinkingMessage.new(type:, hash:)
  when :tool_result
    Messages::ToolResultMessage.new(type:, hash:)
  when :tool_use
    Messages::ToolUseMessage.new(type:, hash:)
  when :user
    Messages::UserMessage.new(type:, hash:)
  else
    Messages::UnknownMessage.new(type:, hash:)
  end
end
```

## Proposed replacement

```ruby
def from_hash(hash)
  type = hash.delete(:type)&.to_sym
  class_name = :"#{type}_message".to_s.camelize  # e.g. :tool_use → "ToolUseMessage"
  klass = if Messages.const_defined?(class_name, false)
    Messages.const_get(class_name, false)
  else
    Messages::UnknownMessage
  end
  klass.new(type:, hash:)
end
```

The `false` argument to `const_defined?`/`const_get` restricts the lookup to `Messages` itself (no ancestor walk), which is the safest approach.

## Why

- **Open/Closed Principle**: Adding a new message type (e.g. `:image`) only requires creating `Messages::ImageMessage` — no need to touch the factory method.
- **Precedent**: This mirrors how `Workflow#load_cog_or_provider` already uses `camelize`/`constantize` for dynamic class resolution (see `lib/roast/workflow.rb:116-119`).
- **Less surface area**: Eliminates 16 lines of mechanical case branches that are easy to forget when adding new types.

## Edge cases to verify

- `type` is `nil` → `"_message".camelize` → `"Message"` — `Messages.const_defined?("Message", false)` returns `false`, so we correctly fall through to `UnknownMessage`.
- Unknown type like `:foo` → `"FooMessage"` — not defined in `Messages`, falls through to `UnknownMessage`. ✓
- Existing types all resolve correctly (confirmed: the naming convention holds for all 8 current types). ✓

## Acceptance criteria

- [ ] Replace the case statement with dynamic resolution
- [ ] All existing tests in `test/roast/cogs/agent/providers/claude/messages/` continue to pass
- [ ] Add a test confirming that an unknown type falls back to `UnknownMessage`

juniper-shopify:
Closes: https://github.com/Shopify/roast/issues/911

### DIFF
--- a/lib/roast/cogs/agent/providers/claude/message.rb
+++ b/lib/roast/cogs/agent/providers/claude/message.rb
@@ -22,26 +22,24 @@ module Roast
               #: (Hash[Symbol, untyped]) -> Message
               def from_hash(hash)
                 type = hash.delete(:type)&.to_sym
-                case type
-                when :assistant
-                  Messages::AssistantMessage.new(type:, hash:)
-                when :result
-                  Messages::ResultMessage.new(type:, hash:)
-                when :system
-                  Messages::SystemMessage.new(type:, hash:)
-                when :text
-                  Messages::TextMessage.new(type:, hash:)
-                when :thinking
-                  Messages::ThinkingMessage.new(type:, hash:)
-                when :tool_result
-                  Messages::ToolResultMessage.new(type:, hash:)
-                when :tool_use
-                  Messages::ToolUseMessage.new(type:, hash:)
-                when :user
-                  Messages::UserMessage.new(type:, hash:)
+                message_class = resolve_message_class(type)
+                message_class.new(type:, hash:)
+              end
+
+              private
+
+              #: (Symbol?) -> singleton(Message)
+              def resolve_message_class(type)
+                return Messages::UnknownMessage if type.nil?
+
+                class_name = "#{type}_message".camelize
+                if Messages.const_defined?(class_name, false)
+                  Messages.const_get(class_name, false) # rubocop:disable Sorbet/ConstantsFromStrings
                 else
-                  Messages::UnknownMessage.new(type:, hash:)
+                  Messages::UnknownMessage
                 end
+              rescue NameError
+                Messages::UnknownMessage
               end
             end
 
--- a/test/roast/cogs/agent/providers/claude/message_test.rb
+++ b/test/roast/cogs/agent/providers/claude/message_test.rb
@@ -149,6 +149,13 @@ module Roast
               assert_kind_of Messages::UnknownMessage, message
             end
 
+            test "from_hash with invalid constant name type creates UnknownMessage" do
+              hash = { type: :"some-hyphenated-type", foo: "bar" }
+              message = Message.from_hash(hash)
+
+              assert_kind_of Messages::UnknownMessage, message
+            end
+
             test "from_hash removes type from hash" do
               hash = { type: :text, text: "Hello" }
               Message.from_hash(hash)


juniper-shopify:
* **#913** <a href="https://app.graphite.com/github/pr/Shopify/roast/913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/cogs/agent/providers/claude/message.rb
```diff
@@ -22,25 +22,21 @@ def from_json(json, raw_dump_file: nil)
...
-                  Messages::ToolUseMessage.new(type:, hash:)
-                when :user
-                  Messages::UserMessage.new(type:, hash:)
+                message_class = resolve_message_class(type)
+                message_class.new(type:, hash:)
+              end
+
+              private
+
+              #: (Symbol?) -> Class
```
LasmarKhalifa: I think changing this to `#: (Symbol?) -> singleton(Message)` would be more precise and fix the failing CI check.

lib/roast/cogs/agent/providers/claude/message.rb
```diff
@@ -22,25 +22,21 @@ def from_json(json, raw_dump_file: nil)
...
+              def resolve_message_class(type)
+                return Messages::UnknownMessage if type.nil?
+
+                class_name = "#{type}_message".camelize
```
LasmarKhalifa, line 35: Are we ever concerned by the fact that `.camelize` could give us a result that generates a `nameError` when evaluated in `Messages.const_defined?(class_name, false)` (such as if the `type` has a hyphen)? 

We can expect that it shouldn't really happen but it might be worth it to add a `rescue nameError`that also returns `Messages::UnknownMessage`for extra robustness.

dersam: This would be a useful thing to write a test to validate.

lib/roast/cogs/agent/providers/claude/message.rb
juniper-shopify: I agree

LasmarKhalifa (approved):
lgtm!

dersam (approved):

PR #915, opened 2026-06-04
[PROTOTYPE] tool_use formatter implementation

LasmarKhalifa:
Playground for tool use implementations. Most files are just for testing, the relevant ones to look at is `tool_use.rb`

A workflow was created to see the tool uses in action.

Things that were considered:

Dual-line tool use: This was a good idea but it made the output harder to understand.  

Single-line without truncation:  
Long tool_use messages would mess up the visual coherence of the terminal output.

single-line with truncation:  
This is what I landed on as my current preferred formatting style. It is easy to understand and takes a single line in most cases.  
Here's what the output looks like from most tools:  
  
Things left to do in this prototype:

- The bash tool can output a long description of the tool. We should look at whether it is worth truncating it or not.
- There are 2 tools that were not in my original list of tools: `taskcreate` and `taskupdate` we need formatters for them too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ coverage
 
 .dev
 .shadowenv
+
+*.log
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+exec ruby -Itest -Ilib "$@"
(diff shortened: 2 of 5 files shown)


LasmarKhalifa:
* **#915** <a href="https://app.graphite.com/github/pr/Shopify/roast/915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #946, opened 2026-06-15
[PROTOTYPE] tool result formatters

LasmarKhalifa:
Prototype for tool_result formatters.   

Example outputs for tool results to illustrate the proposed design:  
￼

We use Nokogiri here, an XML parser to parse the message for TaskOutput, which can look something like:  "\<retrieval_status>success\</retrieval_status>\\n\\n\<task_id>adfgh456b87d647\</task_id>\\n\\n\<task_type>local_agent\</task_type>\\n\\n\<status>completed\</status>\\n\\n\<output>\\nThe total count is 0.\\n\\n**No existing issues found.**\\n\\nThere are no GitHub issues in this org with the `special` label that mention `Class::Special`.\\n\</output>”.    
This allows us to robustly extract the parts we want (retrieval status and status).   

We have 2 versions in this prototype:  
`tool_result.rb` and `tool_result_dry.rb`  The dry version focuses on minimizing repetition and lines of code, while the other one is explicit and localizes all the code for each formatter in its own method.
I think a hybrid approach is best, where we take the `tool_result.rb` but move the error lines in the formatter as well as create a helper method (like the `ok` method in the dry version) for formatters that repeat the same output format, but I’m open to other suggestions!


LasmarKhalifa:
* **#946** <a href="https://app.graphite.com/github/pr/Shopify/roast/946?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/946?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #961, opened 2026-06-17
Add checksums to Gemfile.lock

joshheinrichs-shopify:
This makes it easier for us to package this project with nix.


dersam:
How do we update these checksums on a version bump? Will bundler handle that automatically?

joshheinrichs-shopify:
> How do we update these checksums on a version bump? Will bundler handle that automatically?

Yep!

dersam (approved):

PR #965, opened 2026-06-22
Add internal workflow to maintain up to date LLM model references

### LINKED ISSUE #875: Replace deprecated models in docs and tutorial workflows
Currently, some docs and workflows on Roast use deprecated models (such as "claude-3-5-haiku-20241022"). As new models come out frequently and old ones get deprecated, it would be useful to:

1. Replace current deprecated models with newer ones
2. Have a way to periodically do a model "cleaning" easily (a perfect use case for a Roast workflow!)

LasmarKhalifa:
Closes https://github.com/Shopify/roast/issues/875

One thing to keep in mind while reviewing this: Is it good to update test fixtures too, or should we leave them as is? The current implementation addresses everything, including tests.

### DIFF
--- a/internal/workflows/maintenance/deprecated_models_docs_updater.rb
+++ b/internal/workflows/maintenance/deprecated_models_docs_updater.rb
@@ -1,0 +1,78 @@
+# typed: true
+# frozen_string_literal: true
+
+#: self as Roast::Workflow
+
+# Looks at the Anthropic (https://platform.claude.com/docs/en/about-claude/model-deprecations) and
+# OpenAI (https://developers.openai.com/api/docs/deprecations) model-deprecation pages and updates all
+# docs, doc comments and code references to deprecated/retired models to reflect those changes. This is
+# meant to be run periodically to keep all the references to models up to date.
+
+config do
+  agent do
+    provider :claude
+    model "claude-opus-4-8"
+    quiet!
+  end
+end
+
+execute do
+  agent(:model_verifier) do
+    <<~PROMPT
+      Check both of these pages for deprecated/retired models and their suggested replacements:
+      - Anthropic (Claude): https://platform.claude.com/docs/en/about-claude/model-deprecations
+      - OpenAI: https://developers.openai.com/api/docs/deprecations
+
+      Include every deprecated or retired model that has a recommended replacement, from both pages.
+      Only include models — ignore deprecated API endpoints, tools, or other non-model features.
+
+      Output ONLY a JSON object of exactly this shape, with no surrounding prose:
+      {"outdated": [{"model": "<outdated_model>", "replacement": "<replacement_model>"}]}
+
+      If there are none, output {"outdated": []}.
+    PROMPT
+  end
+
+  agent(:model_finder) do
+    <<~PROMPT
+      You are a code search engine. Search through the codebase and documentation for any references to
+      the outdated models below, including test files, and list every place where they are mentioned.
+
+      The input is a JSON object of the form:
+      {"outdated": [{"model": "<outdated_model>", "replacement": "<replacement_model>"}]}
+
+      INPUT:
+      #{agent!(:model_verifier).response}
+
+      Output ONLY a JSON object of exactly this shape, with no surrounding prose, carrying the
+      replacement through for each model you find a reference to:
+      {"references": [{"model": "<outdated_model>", "replacement": "<replacement_model>", "locations": ["<file_path>:<line_number>"]}]}
+
+      If you find no references, output {"references": []}.
+    PROMPT
+  end
+
+  agent(:updater) do |my|
+    finder = agent!(:model_finder)
+    skip! if finder.json!.fetch(:references, []).empty?
+    my.session = finder.session
+    <<~PROMPT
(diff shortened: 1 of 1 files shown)


LasmarKhalifa:
* **#967** <a href="https://app.graphite.com/github/pr/Shopify/roast/967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#965** <a href="https://app.graphite.com/github/pr/Shopify/roast/965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/workflows/maintenance/deprecated_models_docs_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+      Output ONLY a JSON object of exactly this shape, with no surrounding prose:
+      {"outdated": [{"model": "<outdated_model>", "replacement": "<replacement_model>"}]}
+
+      If there are none, output {"outdated": []}.
+    PROMPT
+  end
+
+  agent(:model_finder) do
+    <<~PROMPT
+      You are a code search engine. Search through the codebase and documentation for any references to
```
juniper-shopify, line 38: Just an overall observation: I typically don't like the "You are a ..." framing in LLM prompts. I don't think it's generally very effective at getting the outcomes you want, *especially* when it's something like "You are an expert in MySQL" or the like -- it doesn't actually make the model more knowledgeable about that topic area, it just makes it more confident, which is usually the opposite of what you want.

In this case, this line pinged my spidey-sense on this point, but I actually think this is a rare good use of the "You are a ..." framing -- by telling the LLM that it is a code search engine (a different kind of *tool*, vs some kind of expert), you're basically restricting its mandate and refocusing it on just doing the searches you want, making it less likely that it will inadvertently try to make changes to the code while searching. 👍🏻

internal/workflows/maintenance/deprecated_models_docs_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+      the outdated models below, including test files, and list every place where they are mentioned.
+
+      The input is a JSON object of the form:
+      {"outdated": [{"model": "<outdated_model>", "replacement": "<replacement_model>"}]}
+
+      INPUT:
+      #{agent!(:model_verifier).response}
+
+      Output ONLY a JSON object of exactly this shape, with no surrounding prose, carrying the
```
juniper-shopify, line 47: A feature I'd ultimately like to implement in Roast is a way to more stricly specify the JSON schema of the response that you want via Anthropic's schema API (which isn't a trivial design challenge, because I'd want writing it to be very fluid and flexible in the Roast workflow definition, whereas schema is tightly structured and pretty explicit).

But, something that Roast does have to force the agent to emit a clean JSON output is single-cog prompt chaining. You can return an array of strings as the prompt instead of a single string, and Roast will run each of those prompts in subsequent turns in the same session, within the same cog invocation. So you could do this:

```
[
  "do <thing> and output a JSON like { <this> }",
  "repeat the JSON object you just produced"
]
```

Roast will give you the output of the final turn as the cog's output, and it's much more likely to be a clean JSON.

That being said, Roast's JSON parser is pretty good at filtering out the part of the response that is the actual JSON and ignoring the surrounding prose, so this isn't necessary all that often, and you might not even need the "with no surrounding prose" line.

No change needed here since it's working and there's nothing wrong with the way you wrote it. I'm just sharing this for context.

internal/workflows/maintenance/deprecated_models_docs_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+      replacement through for each model you find a reference to:
+      {"references": [{"model": "<outdated_model>", "replacement": "<replacement_model>", "locations": ["<file_path>:<line_number>"]}]}
+
+      If you find no references, output {"references": []}.
+    PROMPT
+  end
+
+  agent(:updater) do |my|
+    finder = agent!(:model_finder)
+    skip! if finder.json!.fetch(:references, []).empty?
```
juniper-shopify, line 57:
```suggestion
    skip! if finder.json![:references].blank?
```

this will handle the case where the value is `nil` as well as `[]` (or `""`, etc.)

juniper-shopify (approved):
one tweak to make before merging but otherwise looks good to me

internal/workflows/maintenance/deprecated_models_docs_updater.rb
LasmarKhalifa: Thanks for the info!!

LasmarKhalifa:
### Merge activity

* **Jun 25, 3:16 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/965).
* **Jun 25, 3:16 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/965).

PR #990, opened 2026-06-25
Detect and raise on max-token truncation in chat cog

dersam:
## Problem

The chat cog had no awareness of output token limits. When an LLM response hit the max token limit, the truncated content was silently passed through as complete output. This is a footgun — workflows that depend on the full response would proceed with incomplete data and fail in confusing ways downstream.

The underlying `ruby_llm` gem (v1.15.0) always sends a `max_tokens` value in the request payload (Anthropic uses `model.max_tokens || 4096`, OpenAI uses family-specific limits), but the chat cog never checked whether the response actually hit that ceiling.

## Solution

Add `MaxTokensExceededError` (subclass of `Cog::CogError`) raised after each chat call when `response.output_tokens >= effective_max_tokens`.

**Detection heuristic**: ruby_llm 1.15.0 does not expose `stop_reason`/`finish_reason` from provider responses, so detection is metric-based — if the output token count equals or exceeds the effective max token limit, the content was almost certainly cut off mid-generation.

**Effective max tokens resolution**:
1. Use `chat.model.max_tokens` from model metadata if available
2. Fall back to `4096` for Anthropic (matching the provider's payload behavior)
3. Return `nil` (skip the check) for non-Anthropic providers when the model isn't in the registry (e.g., `assume_model_exists` with an unknown model)

The check is also skipped when `response.output_tokens` is `nil` — we can't detect what we can't measure.

## Changes

- **`lib/roast/cogs/chat.rb`**: `MaxTokensExceededError`, `ANTHROPIC_DEFAULT_MAX_TOKENS` constant, `verify_response_not_truncated!` and `effective_max_tokens` private methods
- **`test/roast/cogs/chat/max_tokens_test.rb`** (new): 11 tests covering exact equality, exceeding, boundary (one-below), Anthropic 4096 fallback, model-specified limits taking precedence, and skip conditions
- **`test/examples/functional/roast_examples_test.rb`**: Updated `stub_ruby_llm_chat` helper to stub `model` on `mock_chat` so the new `effective_max_tokens` call doesn't trigger unexpected invocation errors

## Test Results

```
11 runs, 17 assertions, 0 failures, 0 errors, 0 skips  (max_tokens_test.rb)
30 runs, 252 assertions, 0 failures, 0 errors, 0 skips  (roast_examples_test.rb)
```

### DIFF
--- a/lib/roast/cogs/chat.rb
+++ b/lib/roast/cogs/chat.rb
@@ -20,6 +20,14 @@ module Roast
     #
     # For tasks requiring local filesystem access or locally-configured tools, use the `agent` cog instead.
     class Chat < Cog
+      # Raised when the LLM response hits the max token limit, indicating the output was
+      # truncated and should not be considered usable.
+      class MaxTokensExceededError < Cog::CogError; end
+
+      # Anthropic always sends a max_tokens value in the request payload with this fallback
+      # when the model metadata doesn't specify one (see ruby_llm's Anthropic::Chat#build_base_payload).
+      ANTHROPIC_DEFAULT_MAX_TOKENS = 4096
+
       # The configuration object for this chat cog instance
       #
       #: Roast::Cogs::Chat::Config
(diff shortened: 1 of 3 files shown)


mathiusj (approved):

lib/roast/cogs/chat.rb
```diff
@@ -60,11 +68,53 @@ def execute(input)
...
+      end
+
+      # Determine the effective max token limit for the chat request.
+      #
+      # Returns nil if the limit cannot be determined (e.g., the model is not in the registry
+      # and the provider doesn't have a known default).
+      #
+      #: (RubyLLM::Chat) -> Integer?
+      def effective_max_tokens(chat)
+        max_tokens = chat.model&.max_tokens
```
LasmarKhalifa: I think `effective_max_tokens` ends up being nil in most cases here unless it's an anthropic chat.   
`max_tokens` seems to be dead code that always resolves to `nil`  
  
On line 39-44 we have: 

```ruby
def execute(input)
        chat = ruby_llm_context.chat(
          model: config.valid_model,
          provider: config.valid_provider!,
          assume_model_exists: !config.verify_model_exists?,
        )
```

`verify_model_exists?`  defaults to false: 

```ruby
        def verify_model_exists?
          @values.fetch(:verify_model_exists, false)
        end
```

I can't see any place where `verify_model_exists` is set to true. So `assume_model_exists` ends up being `true`.   
  
In RubyLLM, if we have `assume_model_exists` as true, we end up doing

```ruby
model ||= Model::Info.default(model_id, provider_class.slug)
```

which calls

```ruby
def self.default(model_id, provider)
  new(
    id: model_id,
    name: model_id.tr('-', ' ').capitalize,
    provider: provider,
    capabilities: %w[function_calling streaming vision structured_output],
    modalities: { input: %w[text image], output: %w[text] },
    metadata: { warning: 'Assuming model exists, capabilities may not be accurate' }
  )
end
```

This doesn't contain a `max_output_tokens`, so it will always resolve to nil.   
  
As opposed to the path where `assume_model_exists` is false, which calls 

```ruby
model = Models.find model_id, provider
```

which ends up calling initialize instead of .default, which contains the `max_output_tokens`

```ruby
def initialize(data)
        @id = data[:id]
        @name = data[:name]
        @provider = data[:provider]
        @family = data[:family]
        @created_at = Utils.to_time(data[:created_at])&.utc
        @context_window = data[:context_window]
        @max_output_tokens = data[:max_output_tokens]
        @knowledge_cutoff = Utils.to_date(data[:knowledge_cutoff])
        @modalities = Modalities.new(data[:modalities] || {})
        @capabilities = data[:capabilities] || []
        @pricing = Pricing.new(data[:pricing] || {})
        @metadata = data[:metadata] || {}
      end
```

LasmarKhalifa: Something like

```ruby
  RubyLLM.models.find(config.valid_model, config.valid_provider!)&.max_tokens
rescue RubyLLM::ModelNotFoundError
  nil
```

Could be a good way to get the limits of each provider, without having to touch `assume_model_exists`.|  
We're essentially looking at RubyLLM.models, which loads values from [this registry](https://github.com/crmne/ruby_llm/blob/1.15.0/lib/ruby_llm/models.json), and extracting the value directly.

dersam: Good catch, thanks! PR updated.

LasmarKhalifa (approved):
Looks good! Nice improvement to Roast!

PR #999, opened 2026-06-26
Add env variable override for the default agent provider

LasmarKhalifa:
Closes https://github.com/Shopify/roast/issues/998

Adds ROAST_DEFAULT_AGENT_PROVIDER as an env variable. If a workflow sets a provider, that takes precedence.

The value set in the env var is stripped and downcased, so that `PI`, `pi` and `pi `(with a trailing space) all resolve to `:pi`

PR #1001, opened 2026-06-28
Add ROAST_DEFAULT_CHAT_PROVIDER env variable

### LINKED ISSUE #1000: Allow overriding default chat provider with env variable
## Problem

The `chat` cog defaults to `:openai`, but there is no way to change this globally without editing every workflow file. This creates friction for users who only have Anthropic (or other provider) credentials — the tutorial fails immediately on the first `chat` cog with `no api key provided`.

This is especially jarring because the `agent` cog already defaults to `:claude`, so a user set up for Claude Code can run agent workflows but not chat workflows without per-file config changes.

PR #999 just solved this for the `agent` cog by introducing `ROAST_DEFAULT_AGENT_PROVIDER`. The README in that PR even explicitly notes: _"This affects the `agent` cog only; the `chat` cog's provider is unaffected."_

## Proposed Solution

Add a `ROAST_DEFAULT_CHAT_PROVIDER` environment variable, mirroring `ROAST_DEFAULT_AGENT_PROVIDER`:

```bash
export ROAST_DEFAULT_CHAT_PROVIDER=anthropic
```

Resolution order (same precedence as agent):
1. Provider explicitly set in the workflow `config` block
2. `ROAST_DEFAULT_CHAT_PROVIDER` env var
3. Built-in default (`:openai`)

The value would be normalized (stripped + downcased) and validated against `PROVIDERS.keys`, raising `InvalidConfigError` on an unknown value.

## Impact

- Zero breaking changes — existing workflows that set a provider are unaffected
- Users with Anthropic/Perplexity/Gemini keys can run the tutorial without modifying any workflow files
- Symmetric with `ROAST_DEFAULT_AGENT_PROVIDER` from #999

sntxrr:
Closes #1000

## Summary

Mirrors `ROAST_DEFAULT_AGENT_PROVIDER` (introduced in #999) for the `chat` cog. Users with a non-OpenAI key can now set a single env variable instead of adding `provider` to every workflow `config` block.

```bash
export ROAST_DEFAULT_CHAT_PROVIDER=anthropic
```

**Resolution order** (same precedence as the agent cog):
1. Provider explicitly set in the workflow `config` block
2. `ROAST_DEFAULT_CHAT_PROVIDER` env var (normalized: stripped + downcased)
3. Built-in default (`:openai`)

An unrecognized value raises `InvalidConfigError` at resolution time.

## Changes

- `lib/roast/cogs/chat/config.rb` — add `DEFAULT_PROVIDER_ENV_VAR` constant and update `valid_provider!` to check it
- `test/roast/cogs/chat/config_test.rb` — guard existing default-provider tests against the env var; add 5 new tests covering env var resolution, precedence, normalization, invalid values, and blank values
- `README.md` — rewrite the "Provider Configuration" and "Chat cog" sections to document all four supported providers and both env vars

## Test plan

- [ ] `bundle exec ruby -Itest test/roast/cogs/chat/config_test.rb` — 63 tests, 0 failures
- [ ] Set `ROAST_DEFAULT_CHAT_PROVIDER=anthropic` with `ANTHROPIC_API_KEY` set, run `bin/roast execute tutorial/01_your_first_workflow/hello.rb` — succeeds without a `provider` in the workflow config
- [ ] Set `ROAST_DEFAULT_CHAT_PROVIDER=invalid` — workflow startup raises `InvalidConfigError`
- [ ] Workflow with explicit `provider :openai` in config is unaffected by the env var

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,23 +74,26 @@ gem 'roast-ai'
 
 ## Provider Configuration
 
-Roast provider settings are configured in workflow `config` blocks. There is not currently a CLI flag or environment variable that changes the default provider globally; edit the workflow config to select a different provider.
+Roast provider settings are configured in workflow `config` blocks. To change the default **chat** provider without editing each workflow, set the `ROAST_DEFAULT_CHAT_PROVIDER` environment variable (e.g. `export ROAST_DEFAULT_CHAT_PROVIDER=anthropic`). To change the default **agent** provider, set `ROAST_DEFAULT_AGENT_PROVIDER`. A `provider` set in a workflow's `config` always takes precedence over these variables, and an invalid value raises an error.
 
 ### Chat cog
 
-The `chat` cog currently supports the OpenAI provider. It uses `OPENAI_API_KEY` by default, and `OPENAI_API_BASE` can override the default base URL (`https://api.openai.com/v1`).
+The `chat` cog supports **OpenAI**, **Anthropic**, **Perplexity**, and **Gemini**. It defaults to `:openai` (override globally with `ROAST_DEFAULT_CHAT_PROVIDER`). Each provider reads its API key from a provider-specific environment variable:
 
-```bash
-export OPENAI_API_KEY=...
-```
+| Provider | API key env var | Base URL env var |
+|---|---|---|
+| `:openai` | `OPENAI_API_KEY` | `OPENAI_API_BASE` |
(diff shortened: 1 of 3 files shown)


sntxrr:
I have signed the CLA!

dersam (approved):
Thanks!

PR #1005, opened 2026-06-29
[PROTOTYPE] Pi tool formatters

LasmarKhalifa:
Prototype for Pi tool formatters. Mirrors the current implementation of Claude tool formatters. Pi only exposes 7 tools by default, as stated [here](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/usage.md#tool-options:~:text=Built%2Din%20tools%3A%20read%2C%20bash%2C%20edit%2C%20write%2C%20grep%2C%20find%2C%20ls.).

The final implementation would include doc comments (mirroring the style of the Claude formatters), as well as tests.  
  
Current tool formatting (includes every tool call and result as well as an error case):

Proposed formatting (same tool calls, results and error case):

### DIFF
--- a/lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
+++ b/lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
@@ -74,6 +74,26 @@ module Roast
               end
 
               #: () -> String
+              def format_grep
+                pattern, path, glob = arguments.values_at(:pattern, :path, :glob)
+                base = "GREP \"#{truncate(pattern)}\""
+                base = "#{base} #{path}" if path.present?
+                glob.present? ? "#{base} (glob=#{glob})" : base
+              end
+
+              #: () -> String
+              def format_find
+                pattern, path = arguments.values_at(:pattern, :path)
+                path.present? ? "FIND #{pattern} (in #{path})" : "FIND #{pattern}"
+              end
+
+              #: () -> String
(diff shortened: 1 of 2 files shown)


LasmarKhalifa:
* **#1005** <a href="https://app.graphite.com/github/pr/Shopify/roast/1005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/1005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dersam:
We should put some more thought into how we handle custom tool display, as Pi is almost certain to have those. Calling them "unknown" isn't good enough, as anybody can define a tool in the harness and Roast should not be expected to know anything other than defaults.

LasmarKhalifa:
> We should put some more thought into how we handle custom tool display, as Pi is almost certain to have those. Calling them "unknown" isn't good enough, as anybody can define a tool in the harness and Roast should not be expected to know anything other than defaults.

​I agree. Made changes so that each unknown tool call shows the name of the tool, and then shows each key-value pair with truncation so the output remains concise. This makes it more of a generic tool formatter rather than one that expresses that this tool is unknown, which is a great improvement. Let me know what you think!

dersam:
Nice, I like that much better.

PR #1012, opened 2026-06-30
restructure Pi tool formatting into per-method dispatch with generic fallback

### LINKED ISSUE #1022: Restructure Pi tool formatting into per-method dispatch with generic fallback
Pi should follow the pattern used in the Claude tool formatters and have per-method dispatching.

LasmarKhalifa:
Closes https://github.com/Shopify/roast/issues/1022

Builds the foundation for the stack by replacing the current tool formatter routing to a per-method dispatch.

### DIFF
--- a/lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
+++ b/lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
@@ -25,31 +25,53 @@ module Roast
               def initialize(id:, name:, arguments:)
                 @id = id
                 @name = name
-                @arguments = arguments
+                @arguments = arguments.is_a?(Hash) ? arguments : {}
               end
 
               #: () -> String?
               def format
                 return unless name
 
-                case name.to_s.downcase
-                when "bash"
-                  "BASH #{arguments[:command]}"
-                when "read"
-                  "READ #{arguments[:path]}"
-                when "edit"
-                  "EDIT #{arguments[:path]}"
-                when "write"
-                  "WRITE #{arguments[:path]}"
-                when "grep"
-                  "GREP #{arguments[:pattern]} #{arguments[:path]}"
-                when "find"
-                  "FIND #{arguments[:pattern]} #{arguments[:path]}"
-                when "ls"
-                  "LS #{arguments[:path]}"
-                else
-                  "TOOL [#{name}] #{arguments.inspect}"
-                end
+                format_method_name = "format_#{name.to_s.downcase}".to_sym
+                return send(format_method_name) if respond_to?(format_method_name, true)
+
+                format_unknown
+              rescue StandardError
+                format_unknown
+              end
+
+              # Truncates each value in a formatted tool-call string to keep terminal output concise.
+              TRUNCATE_LIMIT = 45
+
+              private
+
+              # Formats a tool call for which Roast has no dedicated formatter.
+              #
+              # Output: "<NAME> <key>: <value>, ..." – the upcased tool name, then each
+              # argument as "<key>: <inspected value>", ordered shortest pair first so the
+              # most compact arguments stay visible, joined with ", ". Every value is
+              # truncated to TRUNCATE_LIMIT chars so one large argument can't flood the
+              # line; keys are always shown. No arguments renders the bare "<NAME>".
+              #
+              # Examples:
+              #   WEB_SEARCH max_results: 5, query: "ruby pluralize"
+              #   DEPLOY
+              #
+              #: () -> String
+              def format_unknown
+                label = name.to_s.upcase
+                return label if arguments.empty?
+
+                details = arguments.map { |key, value| "#{key}: #{truncate(value.inspect)}" }.sort_by { |s| [s.length, s] }.join(", ")
+                "#{label} #{details}"
+              end
+
+              # Truncates to TRUNCATE_LIMIT chars, appending "..." when cut. nil -> "".
+              #
+              #: (String?) -> String
+              def truncate(str)
+                s = str.to_s
+                s.length > TRUNCATE_LIMIT ? "#{s[0...TRUNCATE_LIMIT - 3]}..." : s
               end
             end
           end
--- a/lib/roast/cogs/agent/providers/pi/messages/tool_result_message.rb
+++ b/lib/roast/cogs/agent/providers/pi/messages/tool_result_message.rb
@@ -30,23 +30,77 @@ module Roast
                 @tool_name = tool_name
                 @content = content
                 @is_error = is_error
+                @name = (tool_name || "unknown").to_s #: String
+                @input = {} #: Hash[Symbol, untyped]
               end
 
               #: (PiInvocation::Context) -> String?
               def format(context)
-                tool_call = context.tool_call(tool_call_id)
-                name = tool_name || tool_call&.name || "unknown"
-                status = is_error ? "ERROR" : "OK"
+                call = context.tool_call(tool_call_id)
+                @name = (tool_name || call&.name || "unknown").to_s
+                @input = call&.arguments || {}
 
-                # Truncate long tool results for progress display
-                c = content
-                display_content = if c && c.length > 200
-                  "#{c[0..197]}..."
-                else
-                  c
-                end
+                return error_line if is_error
 
-                "#{name.upcase} #{status}#{display_content ? " #{display_content}" : ""}"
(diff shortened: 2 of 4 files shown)


LasmarKhalifa:
* **#1021** <a href="https://app.graphite.com/github/pr/Shopify/roast/1021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1020** <a href="https://app.graphite.com/github/pr/Shopify/roast/1020?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1019** <a href="https://app.graphite.com/github/pr/Shopify/roast/1019?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1018** <a href="https://app.graphite.com/github/pr/Shopify/roast/1018?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1017** <a href="https://app.graphite.com/github/pr/Shopify/roast/1017?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1016** <a href="https://app.graphite.com/github/pr/Shopify/roast/1016?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1014** <a href="https://app.graphite.com/github/pr/Shopify/roast/1014?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1012** <a href="https://app.graphite.com/github/pr/Shopify/roast/1012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/1012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
```diff
@@ -32,24 +32,44 @@ def initialize(id:, name:, arguments:)
...
-                when "grep"
-                  "GREP #{arguments[:pattern]} #{arguments[:path]}"
-                when "find"
-                  "FIND #{arguments[:pattern]} #{arguments[:path]}"
-                when "ls"
-                  "LS #{arguments[:path]}"
-                else
-                  "TOOL [#{name}] #{arguments.inspect}"
-                end
+                format_method_name = "format_#{name.to_s.downcase}".to_sym
```
mathiusj, line 35: just out of curiosity, if this pr was to be deployed on it's own would that remove support for all the tools that were previously supported? :thinking_face: If so, I'd be inclined to find a better way to stack this.

LasmarKhalifa: I wouldn't say it removes support. Each tool call would just route to `format_unknown` which outputs the name of the tool in upper case followed by the `key: value` pairs in the arguments. So it comes out to be pretty much the same thing. I would even say it's probably better because some info is currently omitted (such as the `offset` and `limit` in the `read` tool, but routing `read` calls through `format_unknown` would make those useful bits of information show up!

mathiusj (approved):

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
```diff
@@ -32,24 +32,44 @@ def initialize(id:, name:, arguments:)
...
+                return send(format_method_name) if respond_to?(format_method_name, true)
+
+                format_unknown
+              end
+
+              # Truncate long formatted tool-call strings to keep terminal output to generally one line, accounting for logger prefixing.
```
mathiusj: **Line-length contract vs. what the code enforces.** This comment (and its result-side twin) says the goal is to "keep terminal output to generally one line," but `TRUNCATE_LIMIT` only bounds *individual values*, not the composed line. `format_unknown` joins *every* argument (each truncated to 45) with no cap on the number of args, so N args → up to ~45·N chars. Downstream, `format_grep` appends an untruncated path + glob + limit. Either truncate the final composed string, or soften the wording to "each value" so the comment doesn't overpromise a bound the code doesn't keep.

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
```diff
@@ -32,24 +32,44 @@ def initialize(id:, name:, arguments:)
...
+              # most compact arguments stay visible, joined with ", ". Every value is
+              # truncated to TRUNCATE_LIMIT chars so one large argument can't flood the
+              # line; keys are always shown. No arguments renders the bare "<NAME>".
+              #
+              # Examples:
+              #   WEB_SEARCH max_results: 5, query: "ruby pluralize"
+              #   DEPLOY
+              #
+              #: () -> String
+              def format_unknown
```
mathiusj, line 61: **Robustness: dispatch trusts `arguments` is a well-formed Hash.** `format_unknown` calls `arguments.empty?`/`.map`, and downstream formatters do `arguments.values_at(...)` / `arguments[:edits].length`. If Pi ever sends a non-Hash (or nil) `arguments`, or a formatter's typed assumption is violated (e.g. `:edits` as a String), dispatch raises inside the progress logger. Since this is display-only, the highest-leverage fix lives *here* in the foundation: wrap the `send(format_method_name)` in a rescue that falls back to `format_unknown`. That protects all 7 downstream PRs at once and means per-tool reviewers don't have to scrutinize each formatter's arg assumptions. (Normalizing `@arguments` to `{}` when non-Hash in `initialize` is a cheap complement.)

lib/roast/cogs/agent/providers/pi/messages/tool_result_message.rb
```diff
@@ -30,23 +30,75 @@ def initialize(tool_call_id:, tool_name:, content:, is_error:)
...
+              end
+
+              # Renders "<TOOL> ERROR <message>". The content is shown as-is and intentionally not truncated,
+              # preserving the full diagnostic.
+              #
+              # Examples:
+              #   READ ERROR ENOENT: no such file or directory
+              #
+              #: () -> String
+              def error_line
```
mathiusj, line 94: Minor: the outer `.strip` here only does anything when `content` is blank (trimming the trailing space after `ERROR`). Harmless, just slightly redundant given `content.to_s.strip` is already stripped. Could special-case blank content instead, but not worth churn.

mathiusj (commented):
Foundation looks clean — per-method dispatch with `respond_to?(..., true)` and a generic fallback reads well, and the test rewrite is a nice tightening. A few things to consider before the whole stack lands on top of this.

mathiusj (approved):

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
LasmarKhalifa: Added the fallback!

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
LasmarKhalifa: Changed the wording to make it clearer!

lib/roast/cogs/agent/providers/pi/messages/tool_call_message.rb
```diff
@@ -25,31 +25,53 @@ class ToolCallMessage
...
+                label = name.to_s.upcase
+                return label if arguments.empty?
+
+                details = arguments.map { |key, value| "#{key}: #{truncate(value.inspect)}" }.sort_by(&:length).join(", ")
```
dersam: **Non-stable sort in `format_unknown`.** `sort_by(&:length)` is not guaranteed stable in Ruby — when two argument pairs render to the same string length, their relative order is non-deterministic, which could cause flaky terminal output and flaky tests. Fix: `.sort_by { |s| [s.length, s] }` to tie-break on the rendered string.

LasmarKhalifa: Changed it, thanks for flagging this!

dersam (approved):

LasmarKhalifa:
### Merge activity

* **Aug 5, 3:41 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/1012).
* **Aug 5, 3:41 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/1012).

PR #1030, opened 2026-07-07
Add workflow parameter access in config blocks

### LINKED ISSUE #1006: Add access to workflow parameters in config blocks
Currently, parameters being passed to a workflow are only available in cog input blocks. We should allow them to be accessed in config blocks too. 

This would allow for things like selecting a model based on a target or properties of the args (i.e. selecting a more performant model if there are <5 files being passed in a workflow that reviews files for errors).

LasmarKhalifa:
Closes https://github.com/Shopify/roast/issues/1006

Workflow params (`target!`, `targets`, `arg?`, `args`, `kwarg`, `kwarg!`, `kwarg?`, `kwargs`) are currently only available in `execute` blocks. This PR makes it possible to read them inside cog `config` blocks and the `global` block, so cogs can be configured based on how the workflow was invoked.

Example usage:

```ruby
config do
  chat do
    model(arg?(:fast) ? "gpt-5.4-nano" : "gpt-5")
  end
  global do
    abort_on_failure! if arg?(:strict)
  end
end
```

They are not available in the top-level `config do` body, only within the cog / global blocks nested inside it. The accessors mirror the execute-side ones on [`CogInputContext`](https://github.com/Shopify/roast/blob/main/lib/roast/cog_input_manager.rb#L82-L104) with identical behavior.

### DIFF
--- a/lib/roast/config_manager.rb
+++ b/lib/roast/config_manager.rb
@@ -3,15 +3,18 @@
 
 module Roast
   class ConfigManager
+    include WorkflowParamAccessors
+
     class ConfigManagerError < Roast::Error; end
     class ConfigManagerNotPreparedError < ConfigManagerError; end
     class ConfigManagerAlreadyPreparedError < ConfigManagerError; end
     class IllegalCogNameError < ConfigManagerError; end
 
-    #: (Cog::Registry, Array[^() -> void]) -> void
-    def initialize(cog_registry, config_procs)
+    #: (Cog::Registry, Array[^() -> void], WorkflowContext) -> void
+    def initialize(cog_registry, config_procs, workflow_context)
       @cog_registry = cog_registry
       @config_procs = config_procs
+      @workflow_context = workflow_context
       @config_context = ConfigContext.new #: ConfigContext
       @global_config = Cog::Config.new #: Cog::Config
       @general_configs = {} #: Hash[singleton(Cog), Cog::Config]
@@ -117,6 +120,7 @@ module Roast
       # NOTE: Sorbet expects the proc passed to instance_exec to be declared as taking an argument
       # but our cog_config_proc does not get an argument
       cog_config_proc = cog_config_proc #: as ^(untyped) -> void
+      bind_workflow_params(config_object)
       config_object.instance_exec(&cog_config_proc) if cog_config_proc
       nil
     end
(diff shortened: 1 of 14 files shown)


LasmarKhalifa:
* **#1030** <a href="https://app.graphite.com/github/pr/Shopify/roast/1030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/1030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/roast/config_manager.rb
```diff
@@ -134,8 +136,75 @@ def bind_global
...
+
+    #: (Symbol) -> bool
+    def kwarg?(key)
+      @workflow_context.params.kwargs.include?(key)
+    end
+
+    #: () -> Hash[Symbol, String]
+    def kwargs
+      @workflow_context.params.kwargs.dup
+    end
```
juniper-shopify: This is a big chunk of code that is duplicated from `CogInputManager`. Can we abstract it out into a shared module that both classes import?

juniper-shopify (changes requested):
looking great overall, just one architectural improvement I think is worth making

lib/roast/config_manager.rb
LasmarKhalifa: Yes, made the change!

LasmarKhalifa:
> looking great overall, just one architectural improvement I think is worth making

​Made the change! The duplicate code is now abstracted to `workflow_param_accessors.rb.` Thanks for flagging this!

juniper-shopify (approved):
LGTM!

LasmarKhalifa:
### Merge activity

* **Jul 10, 7:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/roast/1030).
* **Jul 10, 8:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Shopify/roast/1030) rebased this pull request as part of a merge.
* **Jul 10, 8:01 PM UTC**: @LasmarKhalifa merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/roast/1030).

Issue #1008, opened 2026-06-29
Add workflow to validate and update docs & doc comments


LasmarKhalifa:
To avoid doc drift, we should have a workflow that verifies each doc and each doc comment for correctness and creates PRs for the ones that need modifications.

LasmarKhalifa:
This could be either a single workflow, or multiple workflows that each do specific tasks (i.e. one workflow for doc comments and one for docs).

We should also have the ability to either analyze the entire code base or a single file/directory.

hiSandog:
The workflow should separate mechanically verifiable drift from subjective documentation review. Code snippets, referenced constants/signatures, links, and generated CLI output can be checked deterministically; prose updates should produce a scoped report/draft that a maintainer approves rather than automatically opening many speculative PRs. Per-file mode plus a changed-API detector would make the workflow practical in CI without repeatedly analyzing the entire repository.

PR #1031, opened 2026-07-10
Add comment audit workflow

LasmarKhalifa:
Part of https://github.com/Shopify/roast/issues/1008

**In a nutshell:** a Roast workflow that guards against doc-comment and inline comment drift. It reads the comments across the codebase (or any passed target), has an agent check each one against the real implementation, and opens a single PR fixing the stale ones. Read-only by default; it opens a PR only when run with `-- apply`.

**How it works** (one cog per phase):

1. **Enumerate** parses each in-scope file with Prism and extracts atomic units: each contiguous `#` prose block sitting above a definition (class, module, method, attr, constant), plus its RBS signature, as well as inline comments.
2. **Verify** fans out across units in parallel, one agent per unit (Claude Sonnet), judging the comment against the actual code and returning a structured verdict (factual? confidence? suggested fix?).
3. **Plan** classifies each verdict: factual, stale-and-auto-fixable (confident with a safe fix), or low-confidence (report only). (Current threshold is set at 0.8, but this should be adjusted over time).
4. **Apply** (only with `-- apply`) splices the fixes in and opens one draft PR on top of the current branch, gated by a check that requires a clean, pushed, even working tree so the PR diff is only the doc fix.
5. **Summary** prints the findings.

Scope defaults to `lib/roast` plus the rbi shims; pass files or directories as targets to narrow it.

The workflow is heavily commented inline, so that a reader can follow the enumeration and classification logic without having to reverse-engineer it.

Example test run scoped to one file (`lib/roast/cog/output.rb`) here: https://app.graphite.com/github/pr/Shopify/roast/1037

### DIFF
--- a/internal/workflows/maintenance/comment_audit/prompts/verify_comment.md.erb
+++ b/internal/workflows/maintenance/comment_audit/prompts/verify_comment.md.erb
@@ -1,0 +1,51 @@
+You are auditing a SINGLE comment in the Roast codebase for factual correctness
+against the real implementation. This is the only comment you are judging right now.
+
+## The comment under review
+
+- File: `<%= unit[:file] %>`
+<% if unit[:kind] == "inline" %>- An inline comment that documents no definition (it describes nearby code, not a method/class/constant)
+<% else %>- Documents: `<%= unit[:symbol] %>` (<%= unit[:kind] %>)
+<% end %>- Source lines <%= unit[:start_line] %>-<%= unit[:end_line] %>
+<% if unit[:signature] %>- Declared RBS signature, type-checked and authoritative (treat as ground truth; correct the prose to match it, never the reverse): `<%= unit[:signature] %>`
+<% end %>
+Current comment text, verbatim (every line is a comment, shown with its original indentation):
+
+<%= unit[:comment_text] %>
+
+## What to do
+
+<% if unit[:kind] == "inline" %>1. Open `<%= unit[:file] %>` and read the code immediately surrounding lines
+   <%= unit[:start_line] %>-<%= unit[:end_line] %> (the statements this comment describes), plus any
+   closely related code needed to judge the claims. Use only the real code in this repository as ground truth.
+<% else %>1. Open `<%= unit[:file] %>` and read the actual implementation of `<%= unit[:symbol] %>`, plus any
+   closely related code (callers, the methods it delegates to, constants it names) needed to judge
+   the claims. Use only the real code in this repository as ground truth.
+<% end %>2. Check every factual claim the comment makes: described behavior, parameters and return values,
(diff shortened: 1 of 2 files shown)


LasmarKhalifa:
* **#1031** <a href="https://app.graphite.com/github/pr/Shopify/roast/1031?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1034](https://github.com/Shopify/roast/pull/1034) <a href="https://app.graphite.com/github/pr/Shopify/roast/1034?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1037](https://github.com/Shopify/roast/pull/1037) <a href="https://app.graphite.com/github/pr/Shopify/roast/1037?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/1031?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/workflows/maintenance/doc_audit/workflow.rb
```diff
@@ -0,0 +1,424 @@
...
+#            plus the user-facing rbi shims.
+#
+# Read-only (dry-run) runs from any branch and audits the working tree as-is. `apply` opens the PR
+# on top of the branch it is run from: from `main` the fix stacks cleanly on `main`; from a feature
+# branch it stacks on that branch. `apply` requires a clean working tree and the current branch to be
+# pushed and even with its origin counterpart, so the PR diff is only the doc fix.
+#
+# Examples:
+#   # dry-run the full default scope (prints findings, opens nothing, writes nothing):
+#   bin/roast execute internal/workflows/doc_audit/workflow.rb
```
dersam: **Usage examples reference wrong file path.** All four `bin/roast execute` examples omit the `maintenance/` segment — they say `internal/workflows/doc_audit/workflow.rb` but the file is at `internal/workflows/maintenance/doc_audit/workflow.rb`. The CLI's `resolve_workflow_path` will fail with "Workflow file not found" if a user copies any of these commands.

LasmarKhalifa: Nice catch, I had originally created the workflow in `/maintenance`, then moved it to `/maintenance/doc_audit`  but forgot to change the examples.

Note: I changed the name of the folder to `comment_audit` to reflect the fact that it does not just audit doc-comments (I updated it to handle inline comments as well as you had a good point [here](https://app.graphite.com/github/pr/Shopify/roast/1034#comment-PRRC_kwDOOW6Stc7VZe2S).

Thanks for the comment!

PR #1034, opened 2026-07-10
Fix 1 stale doc comment(s) across 1 file(s)

LasmarKhalifa:
Automated doc-comment audit on top of `07-10/doc-comment-audit-workflow` (base `a5f0059424`).

### `lib/roast/cog/output.rb`

- **Roast::Cog::Output::WithNumber#normalize_number_string** (confidence 0.9)
  - The comment says the method removes 'currency codes', but the regex `[\s$¢£€¥,_]` only strips whitespace, the currency symbols $¢£€¥, commas, and underscores — no letters are stripped, so alphabetic currency codes like 'USD' or 'EUR' are not removed.
  - evidence: lib/roast/cog/output.rb:255-261 (normalize_number_string implementation, regex on line 257)

### DIFF
--- a/lib/roast/cog/output.rb
+++ b/lib/roast/cog/output.rb
@@ -249,7 +249,7 @@ module Roast
           candidates.compact.uniq
         end
 
-        # Normalize a number string by removing separators and currency codes and validating format
+        # Normalize a number string by removing separators and currency symbols and validating format
         #
         #: (String) -> String?
         def normalize_number_string(raw)


lib/roast/cog/output.rb
```diff
@@ -249,7 +249,7 @@ def extract_number_candidates(input)
           candidates.compact.uniq
         end
 
-        # Normalize a number string by removing separators and currency codes and validating format
+        # Normalize a number string by removing separators and currency symbols and validating format
```
dersam, line 252: **Inline comment on line 256 has the same stale phrase.** The PR correctly fixes the method-level doc comment ("currency codes" → "currency symbols"), but the inline comment inside the method body (`# Remove common digit separators and currency codes`) has the exact same inaccuracy. Both should be updated together.

LasmarKhalifa: I agree. I had originally scoped this to just doc comments, but looking back it makes sense to include inline comments too. I made a few changes to the workflow itself to allow inline comments to be reviewed. I ran the same file on it again and got [this PR which now fully fixes the comments](https://app.graphite.com/github/pr/Shopify/roast/1037)! Noramlly, it would have updated the same PR instead of creating a new one, but I changed the branch naming convention since we support inline comments too, which made the new PR appear to be a new branch.

Thanks for flagging this, I think it's the right call to have inline comments being processed in the same workflow!

LasmarKhalifa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/roast/1034?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1034** <a href="https://app.graphite.com/github/pr/Shopify/roast/1034?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/roast/1034?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1031** <a href="https://app.graphite.com/github/pr/Shopify/roast/1031?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1037](https://github.com/Shopify/roast/pull/1037) <a href="https://app.graphite.com/github/pr/Shopify/roast/1037?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

